PR #7, opened 2020-04-04
Copy all the source code from eventing-operator repo to new repo

houshengbo:
Fixes #

## Proposed Changes

* This PR is based on the commit 3c5fe6ba9159bb0124dea951d69ad4d3fa87be8b of eventing-operator repo.
* We replace /eventing-operator/ with /operator/ to fit the new repo.
* Updating .gitattributes
* Move serving and eventing APIs into the same directory called operator. The reason is that if we keep API with the same group name in different folders, there will be naming conflict, when we generate the client folder.
* Adding the eventing component
* Merge the e2e and upgrade tests of eventing-operator.
* Merge the YAML of eventing-operator under config dir.
* Merge the scripts of eventing-operator under hack dir.

**Release Note**

```release-note

```


aliok:
LOL I don't get why code coverage check failed, coverage for listed files are either the same or they're increased :D

@houshengbo, I wouldn't block the merge because of that. Not sure if GH set up to allow manual merging.

hack/generate-yamls.sh
```diff
@@ -50,7 +50,7 @@ fi
 rm -fr ${YAML_OUTPUT_DIR}/*.yaml
 
 # Generated Knative Operator component YAML files
-readonly SERVING_OPERATOR_YAML=${YAML_OUTPUT_DIR}/serving-operator.yaml
+readonly OPERATOR_YAML=${YAML_OUTPUT_DIR}/operator.yaml
```
evankanderson, line 53: Does this end up being run separately for serving and eventing, or ??

PR #13, opened 2020-04-10
Respect HA Replicas setting in HPAs

houshengbo:
Cherry-pick of 940a4ca47efaa0a65e03d706412ed04d7c3e0ea0 in serving-operator
## Proposed Changes

If the operator has a HighAvailabiliy configuration and the replicas are set > 1, we probably want to adjust all HPAs to have at least as many minReplicas as specified in that HA config. An HPA being present is enough of a signal for us to assume that the component supports a HA configuration.

**Release Note**

```release-note
Adjusts all HPAs to have at least the number of replicas specified by a potential HighAvailability configuration.
```

### DIFF
--- a/pkg/reconciler/knativeserving/common/ha.go
+++ b/pkg/reconciler/knativeserving/common/ha.go
@@ -62,9 +62,26 @@ func HighAvailabilityTransform(instance *servingv1alpha1.KnativeServing, log *za
 			}
 		}
 
+		replicas := int64(instance.Spec.HighAvailability.Replicas)
+
 		// Transform deployments that support HA.
 		if u.GetKind() == "Deployment" && deploymentNames.Has(u.GetName()) {
-			if err := unstructured.SetNestedField(u.Object, int64(instance.Spec.HighAvailability.Replicas), "spec", "replicas"); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/reconciler/knativeserving/common/ha.go
```diff
@@ -62,9 +62,26 @@ func HighAvailabilityTransform(instance *servingv1alpha1.KnativeServing, log *za
...
+
 		// Transform deployments that support HA.
 		if u.GetKind() == "Deployment" && deploymentNames.Has(u.GetName()) {
-			if err := unstructured.SetNestedField(u.Object, int64(instance.Spec.HighAvailability.Replicas), "spec", "replicas"); err != nil {
+			if err := unstructured.SetNestedField(u.Object, replicas, "spec", "replicas"); err != nil {
+				return err
+			}
+		}
+
+		if u.GetKind() == "HorizontalPodAutoscaler" {
```
Cynocracy, line 74: Will all HPAs installed necessarily be ones which need to support HA?

Cynocracy: "An HPA being present is enough of a signal for us to assume that the component supports a HA configuration."

Sorry for missing this! Sounds reasonable to me.

PR #21, opened 2020-04-17
Unify operators into one binary.

markusthoemmes:
## Proposed Changes

At this point, there is very little reason to spawn both operators into their own pods. We can unify them into one binary and run them off of the same caches to reduce overhead by quite a bit.

**Release Note**

```release-note
Unified the operator binaries into one.
```


config/role.yaml
```diff
@@ -207,139 +208,4 @@ aggregationRule:
     # automatically.
     - matchExpressions:
         - {key: eventing.knative.dev/release, operator: Exists}
-rules: [] # Rules are automatically filled in by the controller manager.
----
-kind: ClusterRole
```
Cynocracy, line 212: I think you'll still need this Role (or something like it) in order for the operator to begin bootstrapping.

Did you find it was unnecessary? It would be great if we only needed the aggregated role, but I'm suspicious that we need more than that.

markusthoemmes: The original knative-serving-operator role looked like a superset of this. I honestly haven't really checked that in depth though, I mostly eyeballed it.

Cynocracy: Oh, nice. that's entirely possible.

One change we had to make recently was to allow create/delete/list/watch/update for Jobs, I think you may need to readd that (as without it our upgrade tests were failing)

houshengbo:
@markusthoemmes Generally it looks good to me, except a naming issue.
Since we use one operator deployment instead of two operator deployments, I am wondering how we can smoothly upgrade it.
When users install 0.14.x, two operator deployments exists in their kube. When they install 0.15.0, what is the automation we can give them to remove the old deployments?
We need to make our prow job for upgrade able to check that.
This is what I care regarding this change.

jcrossley3:
> When users install 0.14.x, two operator deployments exists in their kube. When they install 0.15.0, what is the automation we can give them to remove the old deployments?

Good point. We can easily have the operator delete the old deployments when it starts up, similar to how the `deleteObsoleteResources` works.

markusthoemmes:
Sorry, I had to heavily rebase against the version changes. Squashed in that process to ease reviewing again.

markusthoemmes:
Oh  nice... a case of "works on my machine" 😂. Gonna look at getting log output from our tests.

pkg/reconciler/upgrade.go
```diff
@@ -0,0 +1,39 @@
...
+// RemovePreUnified resources removes resources that predate the fusion of the operators
+// into a single binary.
+func RemovePreUnifiedResources(kube kubernetes.Interface, oldName string) error {
+	if err := kube.AppsV1().Deployments(system.Namespace()).Delete(oldName, &metav1.DeleteOptions{}); err != nil && !apierrs.IsNotFound(err) {
+		return fmt.Errorf("failed to remove old deployment %s: %w", oldName, err)
+	}
+	if err := kube.CoreV1().ServiceAccounts(system.Namespace()).Delete(oldName, &metav1.DeleteOptions{}); err != nil && !apierrs.IsNotFound(err) {
+		return fmt.Errorf("failed to remove old serviceaccount %s: %w", oldName, err)
+	}
+
```
houshengbo, line 37: Other resources, like roles(knative-serving-operator-aggregated, knative-serving-operator, knative-serving-eventing-aggregated and knative-eventing-operator), and rolebindings(knative-serving-operator-aggregated, knative-serving-operator, knative-serving-eventing-aggregated and knative-eventing-operator), ca also be removed.
I think I have listed them all.

markusthoemmes: They can not, no. I left them all in. They now just aggregate to a different ServiceAccount.

houshengbo:
@markusthoemmes go-coverage asks for unit test coverage of new added files. I am not sure how difficult/easy it can be added to upgrade.go to remove the deployments and service accounts.

markusthoemmes:
Should be straightforward, I got that on my todo list.

PR #24, opened 2020-04-21
Refreshing the docs

jcrossley3:
Our docs were getting a little dated. Here's my attempt at re-organizing them to be more useful to new users.


docs/configuration.md
```diff
@@ -0,0 +1,314 @@
...
+the `controller` container in the `broker` deployment spec, and the
+value of the `DISPATCHER_IMAGE` env var.
+
+```
+spec:
+  registry:
+    override:
+      autoscaler: docker.io/my-org/autoscaler:v0.13.0
+      broker/controller: docker.io/my-org/broker-controller:v0.13.0
+      DISPATCHER_IMAGE: docker.io/my-org/dispatcher:v0.13.0
```
aliok, line 104: [not a comment on the doc]

Oh boy. This is confusing :(

Not really related to how the documentation is. Documentation is good.
The image override mechanism is confusing. 3 different patterns.

We should've accepted @pmorie 's suggestion of creating typed API for these

aliok:
cc @abrennan89 maybe you're interested

docs/configuration.md
jcrossley3: I agree, fwiw, but I also think the `registry` field is an advanced use case.

PR #28, opened 2020-04-21
[release-0.14] Add configmap support for eventing operator and update the manifest of eventing 

houshengbo:
* Add support of configuring ConfigMaps with CR

* Update the manifest of eventing

Fixes #

## Proposed Changes

*
*
*

**Release Note**

```release-note

```


pkg/reconciler/knativeeventing/common/config_maps.go
```diff
@@ -0,0 +1,52 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	eventingv1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
+)
+
+// ConfigMapTransform updates the ConfigMap with the values specified in operator CR
+func ConfigMapTransform(instance *eventingv1alpha1.KnativeEventing, log *zap.SugaredLogger) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		// Let any config in instance override everything else
+		if u.GetKind() == "ConfigMap" {
+			if data, ok := instance.Spec.Config[u.GetName()[len(`config-`):]]; ok {
```
Cynocracy, line 31: could stripping off the 'config-' ever lead to collisions?

pkg/reconciler/knativeeventing/common/extensions_test.go
```diff
@@ -61,20 +61,20 @@ func TestTransformers(t *testing.T) {
 
 	results, err := platform.Transformers(kubeclient.Get(ctx), ke, logger)
 	assertEqual(t, err, nil)
-	// By default, there are 4 functions.
-	assertEqual(t, len(results), 4)
+	// By default, there are 5 functions.
+	assertEqual(t, len(results), 5)
```
Cynocracy, line 65: This seems a bit fragile, is there any way to test that the functions are doing what we'd expect?

pkg/reconciler/knativeeventing/common/config_maps.go
houshengbo: As long as the CM name starts with "config-", it should be fine.
Luckily, all CMs in knative start with "config-".

pkg/reconciler/knativeeventing/common/extensions_test.go
houshengbo: To me, this only tests the number of functions it calls.
For tests verifying each function, we have test cases for each function.

pkg/reconciler/knativeeventing/common/config_maps.go
Cynocracy: Alright :) Do we have that constraint documented anywhere?

I'd just like it to be formalized if we're depending on it here.

PR #29, opened 2020-04-22
Consolidate transformers for common CRD fields, registry and config

### LINKED ISSUE #26: Remove all duplicate transformer code
We don't want to maintain two separate code branches for the fields common to both `KnativeServing` and `KnativeEventing`. All of these files should be removed:

[pkg/reconciler/knativeeventing/common/config_maps.go](../tree/cf1f02a692de/pkg/reconciler/knativeeventing/common/config_maps.go)
[pkg/reconciler/knativeeventing/common/config_maps_test.go](../tree/cf1f02a692de/pkg/reconciler/knativeeventing/common/config_maps_test.go)
[pkg/reconciler/knativeserving/common/config_maps.go](../tree/cf1f02a692de/pkg/reconciler/knativeserving/common/config_maps.go)

[pkg/reconciler/knativeeventing/common/images.go](../tree/cf1f02a692de/pkg/reconciler/knativeeventing/common/images.go)
[pkg/reconciler/knativeeventing/common/images_test.go](../tree/cf1f02a692de/pkg/reconciler/knativeeventing/common/images_test.go)
[pkg/reconciler/knativeserving/common/images.go](../tree/cf1f02a692de/pkg/reconciler/knativeserving/common/images.go)
[pkg/reconciler/knativeserving/common/images_test.go](../tree/cf1f02a692de/pkg/reconciler/knativeserving/common/images_test.go)

Their consolidated versions should be relocated, but I'm not sure where. How about `pkg/transformers`?

jcrossley3:
Fixes #26

Also factored the common test utilities out of images.go and into a common/testing package.

All tests should remain intact -- eventing had a few more than serving -- now all in a single place.


pkg/reconciler/common/extensions.go
```diff
@@ -21,28 +21,16 @@ import (
...
-		ConfigMapTransform(instance, log),
-		ImageTransform(instance, log),
-		GatewayTransform(instance, log),
-		CustomCertsTransform(instance, log),
-		HighAvailabilityTransform(instance, log),
-		ResourceRequirementsTransform(instance, log),
-	}
+func (platforms Platforms) Transformers(kubeClientSet kubernetes.Interface, slog *zap.SugaredLogger) ([]mf.Transformer, error) {
+	log := slog.Named("extensions")
+	result := []mf.Transformer{}
```
markusthoemmes, line 33:
```suggestion
	result := make([]mf.Transformer, 0, len(platforms))
```

pkg/reconciler/common/testing/util.go
```diff
@@ -0,0 +1,84 @@
...
+	var result = unstructured.Unstructured{}
+	err := scheme.Scheme.Convert(obj, &result, nil)
+	if err != nil {
+		t.Fatalf("Could not create unstructured object: %v, err: %v", result, err)
+	}
+	return result
+}
+
+func AssertEqual(t *testing.T, actual, expected interface{}) {
+	if actual == expected {
```
markusthoemmes, line 73:
```suggestion
        t.Helper()
	if actual == expected {
```

(Same for `AssertDeepEqual` and `MakeUnstructured`)

pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -45,7 +46,6 @@ const (
 )
 
 var (
-	platform    common.Platforms
```
markusthoemmes, line 48: Thanks for killing this!

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -159,13 +160,25 @@ func (r *Reconciler) reconcile(ctx context.Context, ks *servingv1alpha1.KnativeS
 
 // Transform the resources
 func (r *Reconciler) transform(ctx context.Context, instance *servingv1alpha1.KnativeServing) (mf.Manifest, error) {
-	logger := logging.FromContext(ctx)
+	log := logging.FromContext(ctx)
```
markusthoemmes, line 163: nit: I'd prefer if those stayed "logger", they're like this in the upstream at least.

pkg/reconciler/common/extensions.go
jcrossley3: done

pkg/reconciler/common/testing/util.go
jcrossley3: done

pkg/reconciler/knativeserving/knativeserving.go
jcrossley3: done

pkg/reconciler/common/images.go
```diff
@@ -138,18 +153,21 @@ func updateCachingImage(instance *servingv1alpha1.KnativeServing, u *unstructure
...
 	}
 	image.Spec.ImagePullSecrets = addImagePullSecrets(image.Spec.ImagePullSecrets, registry, log)
 	log.Debugw("Finished updating image", "image", image.GetName())
 }
 
-func getNewImage(registry *servingv1alpha1.Registry, containerName string) string {
-	if overrideImage := registry.Override[containerName]; overrideImage != "" {
-		return overrideImage
+func getNewImage(registry *v1alpha1.Registry, containerName, parent string) string {
+	if image, ok := registry.Override[parent+delimiter+containerName]; ok {
```
houshengbo, line 166: This made the operator CR support `/containerName` as valid key when parent is empty. I know it does not seem harmful, but is this what you intentionally to enable?

jcrossley3: It wasn't intentional, but I'm not sure it's worth accounting for since `/` isn't a valid character in container names. At least, that's how I interpret this: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#path-segment-names

PR #33, opened 2020-04-27
Don't rely on in-reconciler state to determine last-in-cluster.

markusthoemmes:
## Proposed Changes

Trying to determine the last resource in the cluster via state kept in the reconciler is racy and error-prone. Ask the API instead.

Deployments should also be removed via the ownerRef. An explicit deletion shouldn't be necessary.

**Release Note**

```release-note
NONE
```

/assign @houshengbo @jcrossley3 @aliok

### DIFF
--- a/pkg/reconciler/knativeeventing/knativeeventing.go
+++ b/pkg/reconciler/knativeeventing/knativeeventing.go
@@ -57,7 +57,6 @@ type Reconciler struct {
 	// kubeClientSet allows us to talk to the k8s for operator APIs
 	operatorClientSet clientset.Interface
 	config            mf.Manifest
-	eventings         sets.String
 	// Platform-specific behavior to affect the transform
 	platform common.Platforms
 }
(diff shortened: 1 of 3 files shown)


pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -246,26 +239,36 @@ func (r *Reconciler) delete(instance *eventingv1alpha1.KnativeEventing) error {
...
-		if err := r.config.Filter(mf.ByKind("Deployment")).Delete(); err != nil {
-			r.Logger.Warn(err, "Error deleting deployments")
+
+	// List all KnativeEventings to determine if cluster-scoped resources should be deleted.
+	kes, err := r.KnativeEventingClientSet.OperatorV1alpha1().KnativeEventings("").List(metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list all KnativeEventings: %w", err)
+	}
+
+	if len(kes.Items) == 1 {
```
jcrossley3, line 251: Curious whether there's also a race condition here. Specifically, if the API server tells us there are 2 while both are being deleted, is it safe to rely on the other's reconcile loop to delete the cluster-scoped resources? I assume so, but just checking.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -270,20 +270,33 @@ func (r *Reconciler) delete(ctx context.Context, instance *servingv1alpha1.Knati
...
-	if len(r.servings) == 0 {
-		if err := r.config.Filter(mf.ByKind("Deployment")).Delete(); err != nil {
+
+	// List all KnativeEventings to determine if cluster-scoped resources should be deleted.
+	kes, err := r.knativeServingClientSet.OperatorV1alpha1().KnativeServings("").List(metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list all KnativeServings: %w", err)
+	}
+
+	if len(kes.Items) == 1 {
```
jcrossley3, line 282: Same question above applies, but also makes me wonder if we can factor this "delete the RBAC last" logic into a shared lib.

markusthoemmes: We probably can, yes. This is part of a larger refactor which I wanted to get through first before consolidating too much.

pkg/reconciler/knativeeventing/knativeeventing.go
markusthoemmes: Hm, you are right. I'll make this a little more bulletproof by checking if all items have a DeletionTimestamp set, i.e.: All resources are being deleted.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -270,20 +270,42 @@ func (r *Reconciler) delete(ctx context.Context, instance *servingv1alpha1.Knati
...
+		if ks.GetDeletionTimestamp().IsZero() {
+			allBeingDeleted = false
+			break
+		}
+	}
+
+	if allBeingDeleted {
+		// If the list has only one element, we're currently reconciling the last one and thus
+		// can go ahead and delete cluster-scoped resources.
+		manifest, err := r.transform(ctx, instance)
```
houshengbo, line 294: Along the line 263 before, you use r.transform(instance), but add ctx as another param here. Which one should be the correct one?

houshengbo: I see. they are two different functions

markusthoemmes: Both will eventually have a ctx once I land a followup of this that will use genreconciler for KnativeEventing as well.

It boils down to logger usage: We should always use the logger in the context, because it has all the... well... context already set to it.

markusthoemmes:
hmpf, rebase mistake.

PR #35, opened 2020-04-27
Remove obsolete finalizer from KnativeServing resources.

markusthoemmes:
## Proposed Changes

The genreconciler machinery adds a finalizer automatically. There's no need to define our own and do our own management of it. We need to make sure to remove it from old resources though.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving.go
+++ b/pkg/reconciler/knativeserving/knativeserving.go
@@ -18,6 +18,7 @@ package knativeserving
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	mf "github.com/manifestival/manifestival"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -252,24 +255,40 @@ func (r *Reconciler) checkDeployments(ctx context.Context, manifest *mf.Manifest
 }
 
 // ensureFinalizer attaches a "delete manifest" finalizer to the instance
-func (r *Reconciler) ensureFinalizer(ctx context.Context, manifest *mf.Manifest, instance *servingv1alpha1.KnativeServing) error {
-	for _, finalizer := range instance.GetFinalizers() {
-		if finalizer == finalizerName {
-			return nil
-		}
+func (r *Reconciler) ensureFinalizerRemoval(_ context.Context, _ *mf.Manifest, instance *servingv1alpha1.KnativeServing) error {
```
houshengbo, line 258: If I understand correctly, the comment of this function should be changed into 
```
ensureFinalizerRemovel ensures that Finalizers in the operator Serving CR is consistent with the ones generated by genreconciler.
```
The `oldFinalizerName = "delete-knative-serving-manifest"` is removed, if it exists in old operator serving CR.

In the `delete` function, we do not even check the finilizers any more.

markusthoemmes: Good catch, changed the comment.

PR #36, opened 2020-04-28
Move metrics reporting into its own event handler.

markusthoemmes:
## Proposed Changes

This breaks the "state keeping" of the KnativeServing reconciler out of the reconciler and pushes it into a stateless event-handler format. Also moves the stats reporter code into a KnativeServing specific package.

**Release Note**

```release-note
NONE
```

/assign @aliok @houshengbo

### DIFF
--- a/pkg/reconciler/knativeserving/controller.go
+++ b/pkg/reconciler/knativeserving/controller.go
@@ -15,6 +15,7 @@ package knativeserving
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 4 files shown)


pkg/reconciler/knativeserving/controller.go
```diff
@@ -92,6 +93,31 @@ func NewControllerWithConfig(cfg *rest.Config) func(context.Context, configmap.W
 			Handler:    controller.HandleAll(impl.EnqueueControllerOf),
 		})
 
+		// Reporting statistics on KnativeServing events.
```
houshengbo, line 96: Does knativeeventing have the following statsReporting?

markusthoemmes: Nope.

PR #39, opened 2020-04-28
Remove unnecessary KnativeServing code and finishing touches.

markusthoemmes:
## Proposed Changes

As the title stays. All the status update code in KnativeServing is redundant because it uses genreconciler which does all that out of the box.

Also a few last finishing touches.

**Release Note**

```release-note
NONE
```

/assign @houshengbo

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving.go
+++ b/pkg/reconciler/knativeserving/knativeserving.go
@@ -24,7 +24,6 @@ import (
 	mf "github.com/manifestival/manifestival"
 	clientset "knative.dev/operator/pkg/client/clientset/versioned"
 
-	"go.uber.org/zap"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -124,35 +109,13 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *servingv1alpha1
...
-	if err != nil {
-		return err
-	}
-	return nil
-}
-
-func (r *Reconciler) reconcile(ctx context.Context, ks *servingv1alpha1.KnativeServing) error {
+func (r *Reconciler) ReconcileKind(ctx context.Context, ks *servingv1alpha1.KnativeServing) pkgreconciler.Event {
 	logger := logging.FromContext(ctx)
+	ks.Status.InitializeConditions()
```
jcrossley3, line 114: You sure we want to do this all the time? It's not going to overwrite anything important?

markusthoemmes: Nope, serving's controllers do the same thing.

PR #46, opened 2020-04-29
Fix malformed license headers.

markusthoemmes:
## Proposed Changes

These are just hideous to look at and license files should be consistent throughout the repo. So this fixes that.

**Release Note**

```release-note
NONE
```

/assign @aliok @houshengbo


aliok:
/lgtm
/approve

I wouldn't normally get any PR merged before members from other vendors see it but this is a simple PR.

PR #49, opened 2020-04-29
Eliminate the wasteful updates due to the aggregated rules

jcrossley3:
For any `ClusterRole` with an `aggregationRule`, we read its rules from the database in the transformer, since the manifest will never contain the correct rules.

### DIFF
--- a/pkg/reconciler/knativeserving/common/aggregated_rules_test.go
+++ b/pkg/reconciler/knativeserving/common/aggregated_rules_test.go
@@ -1,0 +1,114 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	util "knative.dev/operator/pkg/reconciler/common/testing"
+	"sigs.k8s.io/yaml"
+)
+
+type ruleAggregationFixture struct {
+	Name              string
+	Input             *unstructured.Unstructured
+	Database          *unstructured.Unstructured
+	OverwriteExpected bool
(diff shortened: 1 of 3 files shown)


pkg/reconciler/knativeserving/common/aggregated_rules_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	util "knative.dev/operator/pkg/reconciler/common/testing"
+	"sigs.k8s.io/yaml"
+)
+
+type ruleAggregationFixture struct {
+	Name              string
+	Input             *unstructured.Unstructured
+	Database          *unstructured.Unstructured
```
markusthoemmes, line 32: This is usually named „Existing“

pkg/reconciler/knativeserving/common/aggregated_rules_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"testing"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	util "knative.dev/operator/pkg/reconciler/common/testing"
+	"sigs.k8s.io/yaml"
+)
+
+type ruleAggregationFixture struct {
```
markusthoemmes, line 29: Want to inline this into the test?

pkg/reconciler/knativeserving/common/aggregated_rules_test.go
```diff
@@ -0,0 +1,114 @@
...
+		return
+	}
+	for _, test := range tests {
+		t.Run(test.Name, func(t *testing.T) {
+			runRuleAggregationTest(t, &test)
+		})
+	}
+}
+
+func runRuleAggregationTest(t *testing.T, test *ruleAggregationFixture) {
```
markusthoemmes, line 93: Inline?

pkg/reconciler/knativeserving/common/aggregated_rules_test.go
jcrossley3: wdm

pkg/reconciler/knativeserving/common/aggregated_rules_test.go
jcrossley3: sure

pkg/reconciler/knativeserving/common/aggregated_rules_test.go
jcrossley3: you mean inside the `Test...` function?

markusthoemmes: Yep, if you inline the type this gets harder to do and imo is mostly noise as you could just as well read the Test function top to bottom.

Nits though, obviously

PR #54, opened 2020-04-30
[WIP] Layout on how common code might look like using common interfaces.

markusthoemmes:
As the title says, playing with code sharing. Only to determine if this is a valid way to go or if it's too gross.

### DIFF
--- a/pkg/apis/operator/v1alpha1/accessor.go
+++ b/pkg/apis/operator/v1alpha1/accessor.go
@@ -1,0 +1,140 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"knative.dev/pkg/apis"
+)
+
+// Common conditions.
+const (
+	InstallSucceeded     apis.ConditionType = "InstallSucceeded"
+	DeploymentsAvailable apis.ConditionType = "DeploymentsAvailable"
+)
+
+// KComponent is a common interface for accessing meta, spec and status of all known types.
+type KComponent interface {
+	metav1.Object
+	schema.ObjectKind
+
+	GetSpec() KComponentSpec
+	GetStatus() KComponentStatus
+}
+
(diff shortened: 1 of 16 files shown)


pkg/apis/operator/v1alpha1/accessor.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"knative.dev/pkg/apis"
+)
+
+// Common conditions.
+const (
+	InstallSucceeded     apis.ConditionType = "InstallSucceeded"
```
jcrossley3, line 28: I'd prefer for `DependenciesInstalled` to be common. It can be useful for eventing, too. It's essentially a no-op in this operator, but it's a nice hook used by downstream operators.

markusthoemmes: SG, I'll consider that!

pkg/apis/operator/v1alpha1/knativeserving_types.go
```diff
@@ -126,11 +95,6 @@ type KnativeServingSpec struct {
 type KnativeServingStatus struct {
 	duckv1.Status `json:",inline"`
```
jcrossley3, line 96: I can see both eventing and serving having identical status fields, including version, conditions, and possibly `DryRunPatch` if that ever comes to fruition. I know we can account for that through the `Status` interface, but it seems cumbersome to continue replicate the fields in both the `KnativeServingStatus` and `KnativeEventingStatus` structs. Thoughts?

pkg/apis/operator/v1alpha1/accessor.go
```diff
@@ -0,0 +1,82 @@
...
+	DeploymentsAvailable apis.ConditionType = "DeploymentsAvailable"
+)
+
+// Object is a common interface for accessing meta, spec and status of all known types.
+type Object interface {
```
jcrossley3, line 33: As we discussed in slack, I'd like this to be less vague. Here are some suggestions: `Resource`, `CustomResource`, `OperatorInstance`, `OperatorResource`. Maybe `Component` or `KnativeComponent` since I use that word in the docs to describe either eventing or serving. Any of those float your boat?

mattmoor: +1 this caught my eye because there's `runtime.Object` upstream.  We also have `duckv1.KResource`.

pkg/apis/operator/v1alpha1/accessor.go
mattmoor: One thing to consider on Conditions is that we generally try to pick the Type such that "True" conveys success and "False" conveys failure (vs. lack of completion).

@evankanderson has been working to try and upstream some of our thinking into the api-principles, but it's at least the rules we are trying to follow.

pkg/apis/operator/v1alpha1/knativeserving_types.go
markusthoemmes: We can certainly try to reduce boilerplate. In this workstream I'm more interested in collapsing business logic code vs. trying to minimize boilerplate on the typing side though. Imo, reusing business logic is the big value add here.

markusthoemmes:
This has served its purpose.

PR #55, opened 2020-04-30
Simplify deleteObsoleteResources using common helpers.

markusthoemmes:
## Proposed Changes

The `deleteObsoleteResources` functions have become very long behemoths, so this uses a few helpers and a loop to densen these pieces up significantly.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/common/unstructured.go
+++ b/pkg/reconciler/common/unstructured.go
@@ -1,0 +1,36 @@
+/*
+Copyright 2020 The Knative Authors
+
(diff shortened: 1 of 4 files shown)


jcrossley3:
My only problem with this is the PR's title: you can't reduce noise by adding more code than you delete without changing behavior. ;)

markusthoemmes:
True, it locally reduces noise. Damn you boilerplate 😂

PR #57, opened 2020-05-04
Bump ObservedGeneration at the beginning of reconcilation.

markusthoemmes:
## Proposed Changes

`ObservedGeneration` informs the consumer that a reconciler has looked at a rolled out change to `Generation`. This is very useful to determine if a state has already been reconciled by only looking at the respective resource.

For example: If I roll out a change to the spec, initially my status will be "Ready", but Generation will be 2, ObservedGeneration will be 1. I now know that I cannot rely on that Ready state, because my last updated hasn't even been reconciled. I can only rely on it if Generation == ObservedGeneration which is a general pattern.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/knativeeventing/knativeeventing.go
+++ b/pkg/reconciler/knativeeventing/knativeeventing.go
@@ -112,6 +112,7 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *eventingv1alpha
 func (r *Reconciler) ReconcileKind(ctx context.Context, ke *eventingv1alpha1.KnativeEventing) pkgreconciler.Event {
 	logger := logging.FromContext(ctx)
 	ke.Status.InitializeConditions()
+	ke.Status.ObservedGeneration = ke.Generation
 
 	logger.Infow("Reconciling KnativeEventing", "status", ke.Status)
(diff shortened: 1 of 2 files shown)


houshengbo:
@markusthoemmes I am not sure how you are going to use .Status.ObservedGeneration

markusthoemmes:
@houshengbo the general use-case is to be sure that somethign is really "Ready".

For example: If I roll out a change to the spec, initially my status will be "Ready", but Generation will be `2`, `ObservedGeneration` will be `1`. I now know that I cannot rely on that `Ready` state, because my last updated hasn't even been reconciled. I can only rely on it if `Generation == ObservedGeneration` which is a general pattern.

jcrossley3:
You should include that example in the description @markusthoemmes because I had no idea how it would be used either. Thanks for asking @houshengbo

markusthoemmes:
Done, sorry, I assumed it's a known construct.

houshengbo:
@markusthoemmes @jcrossley3 If we want to know the CR is READY, do we need to check (CR.IsReady() && Generation == ObservedGeneration) as a combination in all our test cases to determine if it is ready?

PR #60, opened 2020-05-04
WIP: Add a global predicate to the manifest

houshengbo:
Fixes #

## Proposed Changes

* This PR added a global predicate effective on a reconciler scope. Each time manifest is applied, this predicate is used as filter.
* One use case this PR resolves is that when users specify the custom knative ingress and local gateways, we do not need to install the default gateways.

**Release Note**

```release-note

```

### DIFF
--- a/pkg/reconciler/common/manifestWithPolicy.go
+++ b/pkg/reconciler/common/manifestWithPolicy.go
@@ -1,0 +1,27 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


markusthoemmes:
And how would this global predicate be specified?

jcrossley3:
Does this have any benefit until we change the API to allow the user to choose an ingress? If not, it seems a little like "the tail wagging the dog". In other words, let's let the API drive the structure. Of course, I may be totally misunderstanding the intent here.

houshengbo:
@jcrossley3 No benefits in terms of ingress selection.

I am too dedicated to the current code structure. Currently, we need to use local cluster gateway and knative ingress gateway in the CR, to determine if we use the default gateways or custom gateways.

My first intention is to resolve the scenario that if users use their custom gateways in istio, then we do not need to install local cluster gateway and knative ingress gateway any more. I find it even difficult to convince myself to use a global predicate.

However, let's figure out the ingress selection first. Forget about this change.

PR #62, opened 2020-05-05
Align Eventing conditions with Serving conditions.

markusthoemmes:
## Proposed Changes

One more step towards reusable functions between Eventing and Serving: This aligns the conditions and helpers for setting those conditions between Eventing and Serving.

For now, the helper code is mostly a copy. Potential collapsing of the repetetiveness might come later. This is more about the calling site than it is about the definition site.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/apis/operator/v1alpha1/knativeeventing_lifecycle_test.go
+++ b/pkg/apis/operator/v1alpha1/knativeeventing_lifecycle_test.go
@@ -35,103 +35,103 @@ func TestKnativeEventingGroupVersionKind(t *testing.T) {
 }
 
 func TestKnativeEventingHappyPath(t *testing.T) {
-	ks := &KnativeEventingStatus{}
-	ks.InitializeConditions()
+	ke := &KnativeEventingStatus{}
+	ke.InitializeConditions()
 
-	apistest.CheckConditionOngoing(ks, DependenciesInstalled, t)
-	apistest.CheckConditionOngoing(ks, DeploymentsAvailable, t)
-	apistest.CheckConditionOngoing(ks, InstallSucceeded, t)
+	apistest.CheckConditionOngoing(ke, DependenciesInstalled, t)
+	apistest.CheckConditionOngoing(ke, DeploymentsAvailable, t)
+	apistest.CheckConditionOngoing(ke, InstallSucceeded, t)
 
 	// Install succeeds.
-	ks.MarkInstallSucceeded()
+	ke.MarkInstallSucceeded()
 	// Dependencies are assumed successful too.
-	apistest.CheckConditionSucceeded(ks, DependenciesInstalled, t)
-	apistest.CheckConditionOngoing(ks, DeploymentsAvailable, t)
-	apistest.CheckConditionSucceeded(ks, InstallSucceeded, t)
+	apistest.CheckConditionSucceeded(ke, DependenciesInstalled, t)
+	apistest.CheckConditionOngoing(ke, DeploymentsAvailable, t)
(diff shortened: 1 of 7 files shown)


markusthoemmes:
@jcrossley3 not sure if that's really "possible" with the current scheme. We can have Version on there no problem but I'm not so sure about the condition stuff. Let's do that in a followup?

jcrossley3:
@markusthoemmes as long as you don't name it `Status` which would conflict with the one from the duckv1 package, I think you're good. It will muddy up the literals, of course, but it should be straightforward, no?

markusthoemmes:
@jcrossley3 I think the "issue" is that the conditionSet is a global singleton and not captured by the type itself. The lifecycle functions capture that global conditionSet and do stuff based on it so we'd have to see how we can generalize that.

markusthoemmes:
I guess we could **potentially** use the same conditionset for both, but that'd mean that they can't diverge at all which is not necessarily an assumption I'd like to take right now.

jcrossley3:
I'm absolutely assuming the conditionSet will be identical for both, as well as the other status fields.

markusthoemmes:
I don't. Once we add "Ingress" into Serving, we might want to surface its readiness in a separate condition for example.

jcrossley3:
`DependenciesInstalled` and `DependenciesAvailable` BOOM!

markusthoemmes:
I think that's the wrong condition to use. I thought DependenciesInstalled is for **external** dependencies, i.e.: Handled by an external actor and not by this operator at all.

jcrossley3:
No. I think of it as a placeholder or hook. There's nothing preventing this operator from using it, and I expect it will once ingress selection becomes a part of our API. We never want to do anything that would prevent downstream operators from using it, too, of course.

markusthoemmes:
That will though. It needs to remain untouched by this operator to be usable from a downstream operator. Else they will trample each other's state.

jcrossley3:
They will need to coordinate, that's true.

markusthoemmes:
So, in terms of staging things and merges: I intended to land the interfaces first and not go into too much detail on how best to consolidate code on the definition site.

Once the interfaces are landed, we can separately move definition site consolidation (i.e. making the type definitions more dense) and caller consolidation (reusing business logic) forward.

I think it's valuable for us to move forward, especially looking at business logic reuse. I'm happy to discuss this as part of the definition site consolidation at a later stage. Does that make sense?

pkg/apis/operator/v1alpha1/knativeeventing_lifecycle_test.go
```diff
@@ -44,226 +34,104 @@ func TestKnativeEventingGroupVersionKind(t *testing.T) {
...
-func TestKnativeEventingStatusGetCondition(t *testing.T) {
-	ke := &KnativeEventingStatus{}
-	if a := ke.GetCondition(InstallSucceeded); a != nil {
-		t.Errorf("empty EventingStatus returned %v when expected nil", a)
-	}
-	mc := &apis.Condition{
-		Type:   InstallSucceeded,
-		Status: corev1.ConditionTrue,
+func TestKnativeEventingHappyPath(t *testing.T) {
+	ks := &KnativeEventingStatus{}
```
houshengbo, line 38: We better continue to use the name `ke`

markusthoemmes: Sure, done.

houshengbo:
/lgtm
It looks good to me by aligning Eventing conditions with serving's.

PR #63, opened 2020-05-05
Collapse common spec fields into a CommonSpec.

markusthoemmes:
## Proposed Changes

This forms part of the basis to reduce repetition once we're moving forward with the interfaces proposed in #54.

I wanted to put this one out for discussion early as this is a **breaking change** in the construction of KnativeServing/KnativeEventing resources (see the test changes below). Do we want to go that route? The tradeoff is accepting this change vs. repeating accessor functions for both Eventing and Serving.

**Release Note**

```release-note
NONE
```

/assign @aliok @houshengbo @jcrossley3

### DIFF
--- a/pkg/apis/operator/v1alpha1/common.go
+++ b/pkg/apis/operator/v1alpha1/common.go
@@ -16,7 +16,10 @@ limitations under the License.
 
 package v1alpha1
 
-import "knative.dev/pkg/apis"
+import (
+	corev1 "k8s.io/api/core/v1"
+	"knative.dev/pkg/apis"
+)
 
 const (
 	// DependenciesInstalled is a Condition indicating that potential dependencies have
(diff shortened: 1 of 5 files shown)


test/resources/knativeserving.go
```diff
@@ -118,13 +118,15 @@ func getDeploymentStatus(d *v1.Deployment) corev1.ConditionStatus {
...
 func getTestKSOperatorCRSpec() v1alpha1.KnativeServingSpec {
 	return v1alpha1.KnativeServingSpec{
-		Config: map[string]map[string]string{
-			DefaultsConfigKey: {
-				"revision-timeout-seconds": "200",
-			},
-			LoggingConfigKey: {
-				"loglevel.controller": "debug",
-				"loglevel.autoscaler": "debug",
+		CommonSpec: v1alpha1.CommonSpec{
```
markusthoemmes, line 121: Are we willing to accept this change?

jcrossley3: I'm ok with it. It's only an issue for literals, so just makes tests a little noisier. But I'm willing to trade that for the elimination of the duplicate fields.

pkg/apis/operator/v1alpha1/common.go
```diff
@@ -0,0 +1,69 @@
...
+	// +optional
+	Registry Registry `json:"registry,omitempty"`
+
+	// Override containers' resource requirements
+	// +optional
+	Resources []ResourceRequirementsOverride `json:"resources,omitempty"`
+}
+
+// Config represents a map of upstream config map data.
+type Config map[string]map[string]string
```
houshengbo, line 38: Is this line still useful?

markusthoemmes: good point, I should hook it up with the above.

test/resources/knativeserving.go
houshengbo: @markusthoemmes Do we change the structure in CR yamls?
Based on my understanding, it is no. That is fine.
CommonSpec is included directly in KnativeServingSpec and KnativeEventingSpec. That looks find to me as well.

pkg/apis/operator/v1alpha1/common.go
markusthoemmes: Fixed, took it out.

test/resources/knativeserving.go
markusthoemmes: Yeah, as Jim mentioned this is only about literals in Golang.

markusthoemmes:
Rebased.

aliok:
/approve

This is a very needed change IMO. Thanks for initiating this.

PR #75, opened 2020-05-07
Move install routine into common code and add tests.

markusthoemmes:
## Proposed Changes

As the title says mostly. This factors the install routine into common code and adds unit tests to them. Reuses the now common handler in both reconcilers.

**Release Note**

```release-note
NONE
```

/assign @aliok @jcrossley3 @houshengbo

### DIFF
--- a/pkg/reconciler/common/install_test.go
+++ b/pkg/reconciler/common/install_test.go
@@ -1,0 +1,121 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	mf "github.com/manifestival/manifestival"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"knative.dev/operator/pkg/apis/operator/v1alpha1"
+)
+
+func TestInstall(t *testing.T) {
+	// Resources in the manifest
+	version := "v0.14-test"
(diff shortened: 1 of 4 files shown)


pkg/reconciler/common/install_test.go
```diff
@@ -0,0 +1,121 @@
...
+	roleBinding := NamespacedResource("rbac.authorization.k8s.io/v1", "RoleBinding", "test", "test-role-binding")
+	clusterRole := ClusterScopedResource("rbac.authorization.k8s.io/v1", "ClusterRole", "test-cluster-role")
+	clusterRoleBinding := ClusterScopedResource("rbac.authorization.k8s.io/v1", "ClusterRoleBinding", "test-cluster-role-binding")
+
+	// Deliberately mixing the order in the manifest.
+	in := []unstructured.Unstructured{*deployment, *role, *roleBinding, *clusterRole, *clusterRoleBinding}
+	// Expect things to be applied in order.
+	want := []*unstructured.Unstructured{role, clusterRole, roleBinding, clusterRoleBinding, deployment}
+
+	client := &fakeClient{}
```
jcrossley3, line 44: I should publish a `mf.fake.Client`, I think. I've wanted one more than once.

pkg/reconciler/common/install_test.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+func TestInstall(t *testing.T) {
+	// Resources in the manifest
+	version := "v0.14-test"
+	deployment := NamespacedResource("apps/v1", "Deployment", "test", "test-deployment")
+	role := NamespacedResource("rbac.authorization.k8s.io/v1", "Role", "test", "test-role")
+	roleBinding := NamespacedResource("rbac.authorization.k8s.io/v1", "RoleBinding", "test", "test-role-binding")
+	clusterRole := ClusterScopedResource("rbac.authorization.k8s.io/v1", "ClusterRole", "test-cluster-role")
+	clusterRoleBinding := ClusterScopedResource("rbac.authorization.k8s.io/v1", "ClusterRoleBinding", "test-cluster-role-binding")
```
jcrossley3, line 37: If this ever became unwieldy, I might toss a yaml file beneath `testdata/`

markusthoemmes: I like to have them available programmatically a ton more I must say. As in this case, I can reuse the same objects to create the "in" and "want" lists and even reuse their memory adresses.

pkg/reconciler/common/install_test.go
markusthoemmes: Not quite sure if that's necessary. A little helper that makes me only implement the functions really under test would be nice maybe.

PR #77, opened 2020-05-07
Fix job image override

### LINKED ISSUE #76: Unable to override image for the upgrade job
**Describe the bug**
Unable to override image for the upgrade job.

**Expected behavior**
Able to override image for the upgrade job

**To Reproduce**
No way of overriding jobs in https://github.com/knative-sandbox/operator/blob/master/pkg/reconciler/common/images.go#L46

**Knative release version**
0.14

aliok:
Fixes #76 

## Proposed Changes

* Fix image override for jobs (eg. 0.14 upgrade job for eventing)

**Release Note**

```release-note
Fixed image override for Knative Eventing 0.14 upgrade job
```

### DIFF
--- a/pkg/reconciler/common/images_test.go
+++ b/pkg/reconciler/common/images_test.go
@@ -282,36 +282,36 @@ func runResourceTransformTest(t *testing.T, tt *updateImageTest) {
 	unstructuredDeployment := util.MakeUnstructured(t, util.MakeDeployment(tt.name, corev1.PodSpec{Containers: tt.containers}))
 	deploymentTransform := ImageTransform(&tt.registry, log)
 	deploymentTransform(&unstructuredDeployment)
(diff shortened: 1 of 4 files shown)


pkg/reconciler/common/images_test.go
```diff
@@ -304,6 +311,13 @@ func validateUnstructedDaemonSetChanged(t *testing.T, tt *updateImageTest, u *un
 	util.AssertDeepEqual(t, daemonSet.Spec.Template.Spec.Containers, tt.expected)
 }
 
+func validateUnstructedJobChanged(t *testing.T, tt *updateImageTest, u *unstructured.Unstructured) {
```
houshengbo, line 314: @aliok Would you mind doing a search for `unstructed` across all files. I think we should use `Unstructured`.

aliok: Good catch :) I did a search&replace now in the entire codebase

PR #82, opened 2020-05-08
Override images in PodSpecables

aliok:
Fixes #

## Proposed Changes

*
*
*

**Release Note**

```release-note

```

### DIFF
--- a/pkg/reconciler/common/images.go
+++ b/pkg/reconciler/common/images.go
@@ -21,14 +21,13 @@ import (
 
 	mf "github.com/manifestival/manifestival"
 	"go.uber.org/zap"
-	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/common/images.go
```diff
@@ -45,13 +46,12 @@ var (
 func ImageTransform(registry *v1alpha1.Registry, log *zap.SugaredLogger) mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
 		switch u.GetKind() {
-		// TODO need to use PodSpecable duck type in order to remove duplicates of deployment, daemonSet
 		case "Deployment":
```
aliok, line 49: Not sure if I should check if the type implements `WithPod` or if we should check the `kind` manually.
Left that part as is for now.

PR #83, opened 2020-05-08
Remove metric reporting in KnativeServing.

markusthoemmes:
## Proposed Changes

These metrics don't seem very useful and it's doubtable anybody uses them.

/hold

Pending discussion if we want to ditch this.

**Release Note**

```release-note
Removed metric reporting of KnativeServing changes.
```

/assign @houshengbo @aliok @jcrossley3


jcrossley3:
Looks like they were introduced here: https://github.com/knative/serving-operator/pull/233

houshengbo:
@trshafer @anniefu @garron @jihuin
What is the future plan for knative/serving-operator#233? How do we consume the metrics?

markusthoemmes:
IMO the question is not "how" but "what do we expect them to tell us".

anniefu:
I believe the goal with these metric are to give a sense of how much meaningful work the operator is actually doing, @jihuin might be able to add more clarification.

@houshengbo this uses the knative/pkg metrics pkg, so consuming metrics is the same as knative-serving, uses config-observability.yaml to determine metrics backend.



In general, Knative has a lot of instrumented metrics that I'm sure no one uses. I'm not sure if removing them is a good idea though. The fact that they were placed there means someone might be using them, and I don't think there's a policy for deprecating metrics right now.

We do have the `metrics.backend-destination: none` option in config-observability.yaml now to silence metrics. With the upcoming [Knative metrics overhaul](https://github.com/knative/pkg/blob/master/metrics/README.md), I'm working on providing a spot to filter and drop metrics with the Otel Collector Service's [filter processor](https://github.com/open-telemetry/opentelemetry-collector/issues/560).

markusthoemmes:
@anniefu I'd argue that the metrics we have here are not a great indicator of how much meaningful work the operator is actually doing. I'm happy to take this into the WG for a broader general discussion though. I'd love to get a more clear view on why these have been added in the first place.

jihuin:
The metrics were introduced in response to https://github.com/knative/serving-operator/issues/192, which records that users configure the knative serving via installing/editing/deleting KnativeServing CR.

It's a good question: "what do we expect the metrics to tell us", and I don't have a better answer. What could be an appropriate indicator of the work operator is doing, if not the current one? It would be great to have a broader discussion in WG.

trshafer:
I don't personally have a strong opinion. They were originally made to emit similar reconcile metrics to the knative controller.

houshengbo:
@markusthoemmes @anniefu @jihuin @trshafer 
Let's say we remove this part for now, since it is not quite useful as we expected before.
If there is further need for this feature, we can reconsider the design.
Is everyone ok with it?

markusthoemmes:
Rebased and reran relevant scripts, this should be good to go if you still want to merge it @houshengbo

PR #88, opened 2020-05-11
Fix requirement against openshift/api.

markusthoemmes:
## Proposed Changes

`./hack/update-deps.sh` is broken for me without this. openshift/api doesn't have published versions so that seems to trip `go mod`.

Example:

```console
$ ./hack/update-deps.sh 
warning: ignoring symlink /home/mthoemme/dev/knative/operator/test/config/serving
go: github.com/operator-framework/operator-sdk@v0.15.1 requires
	github.com/operator-framework/operator-lifecycle-manager@v0.0.0-20191115003340-16619cd27fa5 requires
	github.com/openshift/api@v3.9.1-0.20190924102528-32369d4db2ad+incompatible: invalid pseudo-version: preceding tag (v3.9.0) not found
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,11 @@ require (
 
 replace (
 	github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309
+
+	// Required indirectly by manifestival.
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20190924102528-32369d4db2ad
+
+	// Aligning all Kubernetes dependencies.
 	k8s.io/api => k8s.io/api v0.16.4
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.16.4
 	k8s.io/apimachinery => k8s.io/apimachinery v0.16.4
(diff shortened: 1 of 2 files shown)


houshengbo:
@markusthoemmes 
Why do we use openshift/api built based the commit https://github.com/openshift/api/commit/32369d4db2ada1c1cdb95b99b568a9925e2a4480, which is almost 8-month old?
Not using recent commit?

go.mod
```diff
@@ -30,6 +30,11 @@ require (
 
 replace (
 	github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309
+
+	// Required indirectly by manifestival.
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20190924102528-32369d4db2ad
```
jcrossley3, line 35: This surprises me. I would consider this a manifestival bug.

markusthoemmes: I would've looked at that next, but I wanted to fix the immediate bug of not being able to run update-deps at all :joy:

markusthoemmes: Could also be an artifact of manifestival using dep and the operator now using go mod.

markusthoemmes: Okay, I dug it out. Proper fix in https://github.com/manifestival/client-go-client/pull/2 and #93

markusthoemmes:
No longer necessary.

PR #91, opened 2020-05-12
Use :nonroot image for the operator.

markusthoemmes:
## Proposed Changes

Current `ko` prints a fat warning that the default image will be switched soon. This adopts that new default image and suppresses the warning. See https://github.com/google/ko/issues/160.

**Release Note**

```release-note
NONE
```


houshengbo:
stuck with testing again?

PR #92, opened 2020-05-13
[WIP] Prototype for an Ingress API in the operator.

markusthoemmes:
## Proposed Changes

This is just to outline how the changes to the operator would look like. I opted for the "filtering" approach we discussed in the meeting since that was the quickest to implement. If we decide to load the ingresses from distinct files, that's fairly straightforward too though.

**Release Note**

```release-note
NONE
```


pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -117,21 +119,32 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *servingv1alpha1.Knat
...
+	if err != nil {
+		return mf.Manifest{}, err
+	}
 
+	// TODO: This is is a little backwards. We might wanna fix this in manifestival.
+	combined, err := mf.ManifestFrom(mf.Slice(append(r.config.Resources(), ingressManifest.Resources()...)),
+		mf.UseClient(r.config.Client))
+	if err != nil {
+		return mf.Manifest{}, err
+	}
```
jcrossley3, line 133: @trshafer here's a good argument for adding an `Add` function to the interface, if only to keep from having to pass the client and logger options along. https://github.com/manifestival/manifestival/issues/41

trshafer: I was using the readme example yesterday and it took me a while to realize that I needed to pass the client and there was no way to pass the logger. It was purely for debugging as one of the unstructured wasn't properly transforming, so I wanted to pull the resources out and apply them one by one with detailed debugging. I could see `manifest.NewManifest(Slice)`.

markusthoemmes: @jcrossley3 I restructured the code a bit. Arguably it makes more sense now, as the mfClient isn't only used in the manifest's context today (it's also used for a transformer).

As we're in parallel moving towards decoupling operator and operand version, the reconcile loop will have to do the loading of the core manifest anyway. Moving it there now kinda solves the awkwardness of having to carry over client and logger as we're just building it there. Please take another look.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -41,14 +45,24 @@ const (
...
 
+var servingManifest mf.Manifest
+
+func init() {
+	manifest, err := mf.ManifestFrom(mf.Recursive(filepath.Join(os.Getenv("KO_DATA_PATH"), "knative-serving/")))
+	if err != nil {
+		panic(err)
+	}
+	servingManifest = manifest
+}
```
jcrossley3, line 56: This will have to move to `ReconcileKind` eventually, right? Might as well stick it there now. The potential `panic` bothers me.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -113,25 +128,36 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *servingv1alpha1.Knat
...
-	logger.Debug("Transforming manifest")
+	logger.Debug("Loading ingress manifests")
+	ingressManifest, err := ingress.Manifest(ctx, instance)
+	if err != nil {
+		return mf.Manifest{}, err
+	}
 
+	combined, err := mf.ManifestFrom(mf.Slice(append(servingManifest.Resources(), ingressManifest.Resources()...)),
+		mf.UseClient(r.mfClient),
+		mf.UseLogger(zapr.NewLogger(logger.Desugar()).WithName("manifestival")))
```
jcrossley3, line 143: Much nicer, but I still think `manifest.Append` is justified.

pkg/reconciler/knativeserving/knativeserving.go
markusthoemmes: We definitely need to cache things and I didn't want to necessarily bite that off here. I benchmarked the reading a bit and it'll be hard to improve so I think we'll need an in-mem cache.
Before we go full dynamic, I figured I leave it to a future thingy.

jcrossley3: Caching aside, I just meant that since we expect the spec to include a version, we'll need to construct the core manifest in the reconcile loop rather than at startup.

markusthoemmes: Yes, but not as part of the work I'm doing. In my work stream, things are static hence I went with this approach.
The failure modes are the same as today. We logger.Fatal out today, which is equal to a panic.

markusthoemmes: I do agree though, that `init` is mega ugly. I'll bite of a simple caching mechanism as a separate change of this that we can land independently!

jcrossley3: Well then why move the manifest loading into the reconciler in this PR at all?!?!

Not trying to put you on the defensive, just seems weird to use `init` for this. I'll yield to other reviewers, though.

markusthoemmes: Mostly consistency with the ingress bits. I agree it's gross though. I'll work on it, thanks for the feedback! :)

pkg/apis/operator/v1alpha1/knativeserving_types.go
```diff
@@ -53,9 +53,11 @@ type KnativeServingSpec struct {
 	CommonSpec `json:",inline"`
 
 	// A means to override the knative-ingress-gateway
+	// Deprecated: Use Ingress.Istio.KnativeIngressGateway instead.
```
houshengbo, line 56: I agree with the plan of deprecation. Just suggest we keep them at least till 0.16.
Hope in 15, we have KnativeIngressGateway and ClusterLocalGateway in here and Ingress.
For 16 maybe, we use conversion webhook to transform old CR to new CR.

houshengbo:
does retest work?
/retest

cmd/operator/kodata/net-kourier/kourier.yaml
```diff
@@ -0,0 +1,336 @@
...
+                port_value: 18000
+          http2_protocol_options: {}
+          type: STRICT_DNS
+    admin:
+      access_log_path: "/dev/stdout"
+      address:
+        pipe:
+          path: /tmp/envoy.admin
+
+---
```
mattmoor, line 336: Add trailing newline:
```suggestion
---

```

cmd/operator/kodata/knative-serving/0.14.0/4-net-istio.yaml
```diff
@@ -448,4 +453,4 @@ spec:
   selector:
     app: istio-webhook
 
----
+---
```
mattmoor, line 456: Add trailing newline:
```suggestion
---

```

cmd/operator/kodata/knative-serving/4-net-kourier.yaml
```diff
@@ -0,0 +1,336 @@
...
+                port_value: 18000
+          http2_protocol_options: {}
+          type: STRICT_DNS
+    admin:
+      access_log_path: "/dev/stdout"
+      address:
+        pipe:
+          path: /tmp/envoy.admin
+
+---
```
mattmoor, line 336: Add trailing newline:
```suggestion
---

```

cmd/operator/kodata/knative-serving/0.14.0/4-net-istio.yaml
mattmoor, line 456: Add trailing newline:
```suggestion
---

```

cmd/operator/kodata/knative-serving/0.15.1/4-net-kourier.yaml
```diff
@@ -0,0 +1,336 @@
...
+                port_value: 18000
+          http2_protocol_options: {}
+          type: STRICT_DNS
+    admin:
+      access_log_path: "/dev/stdout"
+      address:
+        pipe:
+          path: /tmp/envoy.admin
+
+---
```
mattmoor, line 336: Add trailing newline:
```suggestion
---

```

cmd/operator/kodata/knative-serving/0.15.0/4-net-kourier.yaml
```diff
@@ -0,0 +1,340 @@
...
+                port_value: 18000
+          http2_protocol_options: {}
+          type: STRICT_DNS
+    admin:
+      access_log_path: "/dev/stdout"
+      address:
+        pipe:
+          path: /tmp/envoy.admin
+
+---
```
mattmoor, line 340: Add trailing newline:
```suggestion
---

```

cmd/operator/kodata/knative-serving/0.15.1/4-net-kourier.yaml
```diff
@@ -0,0 +1,340 @@
...
```
mattmoor, line 340: Add trailing newline:
```suggestion
---

```

cmd/operator/kodata/knative-serving/0.15.0/4-net-kourier.yaml
mattmoor, line 340: Add trailing newline:
```suggestion
---

```

cmd/operator/kodata/knative-serving/0.15.1/4-net-kourier.yaml
mattmoor, line 340: Add trailing newline:
```suggestion
---

```

pkg/reconciler/knativeserving/ingress/ingress.go
```diff
@@ -0,0 +1,67 @@
...
+			return true
+		}
+
+		return provider == name
+	}
+}
+
+func Filters(ks *v1alpha1.KnativeServing) mf.Predicate {
+	if ks.Spec.Ingress == nil {
+		return mf.Any(istioFilter)
```
jcrossley3, line 42: Isn't the `Any` redundant? Just return `istioFilter`?

test/upgrade/servingoperator_postupgrade_test.go
```diff
@@ -55,7 +56,17 @@ func TestKnativeServingPostUpgrade(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("Failed to get the manifest for Knative: %v", err)
+		}
+		targetManifest = targetManifest.Filter(ingress.Filters(&v1alpha1.KnativeServing{}))
+
+		expectedDeployments := []string{}
+		for _, resource := range targetManifest.Filter(mf.ByKind("Deployment")).Resources() {
+			expectedDeployments = append(expectedDeployments, resource.GetName())
+		}
+
```
jcrossley3, line 69: I'm not sure what's going on here

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -121,6 +122,9 @@ func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, comp
 		ksc.AggregationRuleTransform(manifest.Client),
 	}
 	extra = append(extra, r.extension.Transformers(instance)...)
+	extra = append(extra, ingress.Transformers(ctx, instance)...)
+
+	*manifest = manifest.Filter(ingress.Filters(instance))
```
jcrossley3, line 127: I was expecting the filtering to be in its own function, independent of the transforms, even if it's a one-liner, as a model of how other API's might be staged.

test/upgrade/servingoperator_postupgrade_test.go
markusthoemmes: I need to add the filters to the manifest before checking the deployments. Otherwise it has the kourier deployments too and complains.

pkg/reconciler/knativeserving/knativeserving.go
markusthoemmes: will do in the proper version, thanks!

markusthoemmes:
@jcrossley3 That's actually a good question :thinking:. The net-* repos have their "own" release cadence, in that they won't get a 0.15.0 just because serving gets one. I don't have a good answer on how to solve that off the top of my head though. They definitely get a minor release in accordance with every serving version minor.

markusthoemmes:
Oh I think I do have an answer :joy:: I accidentally did the right thing. We should probably always ship the latest net-XXX manifests when packaging a new version of serving as I kinda did?

cmd/operator/kodata/knative-serving/0.15.1/4-net-kourier.yaml
```diff
@@ -0,0 +1,340 @@
+apiVersion: v1
```
houshengbo, line 1: Where is the good place to save all the network ingresses?
Shall we separate them from the knative-serving?

houshengbo: Just saw the comments from Jim regarding this aspect.

houshengbo: For one version of operator, do we only associate with one version of network ingress? :-(

markusthoemmes: Good questions and I don't have a good answer to them necessarily. We've discussed in the meetings that we go with a filter-based approach but the comments feel like we actually want a version per ingress? I guess that'd change everything.

markusthoemmes:
I think in light of the `version` work, this will need a serious rethink. I'm releasing my lock on the matter for now since I'll be on vacation soon too.

PR #103, opened 2020-05-28
WIP: Get the latest patch release based on a provided SemVer

houshengbo:
@jcrossley3 @markusthoemmes @evankanderson 

I implement a function called `GetLatestOnlinePatchRelease`, which is able to retrieve the latest PATCH release, based on a given version, org and repo name.

For example, if we would like to get the latest patch release of 0.13.x for knative serving, call `GetLatestOnlinePatchRelease("knative", "serving", "0.13.0")` to get the latest patch version, which is 0.13.3.

We need network connection anyway to install the yamls, which are NOT bundled under kodata, so it does not hurt that we access the releases page online to get the available releases.

With each release of operator, we know a DEFAULT version, which is aligned with other knative components. We can get this version number by reading the kodata structure as in https://github.com/knative/operator/pull/102. We can use this DEFAULT version as the param of `GetLatestOnlinePatchRelease` to fetch the latest PATCH release.

For the `getTargetManifest`, if spec.version is empty, we can get this latest patch version online, try to install it online. If it fails, targetManifest falls back to the currentManifest.

How does this plan sound?


pkg/reconciler/common/read_releases.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if len(releases) == 0 {
+		return releaseTags, nil
+	}
+
+	for _, val := range releases {
+		releaseTag := val.GetTagName()
+		// Remove the first letter v to get MAJOR.MINOR.PATCH
+		release := string(releaseTag[1:len(releaseTag)])
```
mattmoor, line 77: Format Go code:
```suggestion
		release := string(releaseTag[1:])
```

pkg/reconciler/common/read_releases_test.go
```diff
@@ -0,0 +1,31 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
```
mattmoor, line 19: Format Go code:
```suggestion
import (
	"testing"

```

pkg/reconciler/common/read_releases_test.go
```diff
@@ -0,0 +1,31 @@
...
+	util "knative.dev/operator/pkg/reconciler/common/testing"
+	"testing"
```
mattmoor, line 21: Format Go code:
```suggestion

```

pkg/reconciler/common/read_releases.go
mattmoor, line 77: Format Go code:
```suggestion
		release := string(releaseTag[1:])
```

pkg/reconciler/common/read_releases_test.go
mattmoor, line 19: Format Go code:
```suggestion
import (
	"testing"

```

pkg/reconciler/common/read_releases_test.go
mattmoor, line 21: Format Go code:
```suggestion

```

houshengbo:
I updated the doc: https://docs.google.com/document/d/1ehd5Yx4aSottuv7FYW4JeplNULJCs7nnDfy29y3NH8k/edit#heading=h.5502wzf6mqfe

jcrossley3:
Any particular reason you chose the blang library? Seems like this one is much smaller and a better fit for what you're doing, e.g. `Compare` and `MajorMinor`: https://pkg.go.dev/golang.org/x/mod/semver

I commented in the doc, but I'll repeat here: I don't think we should access anything over the network unless `spec.version` is set.

houshengbo:
Drop this PR, since we do not want to access network for empty spec.version.

PR #125, opened 2020-06-03
Create the expected deployments based on version

### LINKED ISSUE #111: Upgrade test suites are rigid and hard-coded with lists of deployment resources
**Describe the bug**
It is very inconvenient to manually check the resources each time, once there is a new release. Knative operator needs to re-check or even re-write some hard coded resources for check-up in upgrade test suites. For example, the list of expected deployment resources. 

**Expected behavior**
The upgrade tests for both serving and eventing should be more dynamic, without hard code of any specific resources. After this issue https://github.com/knative/operator/issues/109 is resolved, operator is able to compare the diffs between two manifests. We can make the upgrade test suites more intelligent.

This improvement is foundational work of the feature https://github.com/knative/operator/issues/56, in terms of test cases.
**To Reproduce**
Every time we release a new version, we need to compare manually the diffs between the current released manifest and the previous one. Find out all of them, and re-implement the resource verification in the upgrade tests.

**Knative release version**
All.

houshengbo:
Fixes #111

## Proposed Changes

* This PR created the expected deployments based on the provided versions
* The hard-coded resources are removed, so the test cases can create the expected deployment resources to verify based on different versions.

**Release Note**

```release-note

```

### DIFF
--- a/test/resources/knativeresources.go
+++ b/test/resources/knativeresources.go
@@ -89,6 +89,8 @@ func IsKnativeDeploymentReady(dpList *v1.DeploymentList, expectedDeployments []s
 	return true, nil
 }
 
+// GetExpectedDeployments will return an array of deployment resources based on the version for the knative
+// component.
 func GetExpectedDeployments(t *testing.T, version, kcomponent string) []string {
 	SetKodataDir()
 	defer os.Unsetenv(common.KoEnvKey)
@@ -105,6 +107,7 @@ func GetExpectedDeployments(t *testing.T, version, kcomponent string) []string {
 	return removeDuplications(deployments)
 }
 
+// SetKodataDir will set the env var KO_DATA_PATH into the path of the kodata of this repository.
 func SetKodataDir() {
 	_, b, _, _ := runtime.Caller(0)
 	koPath := filepath.Join(filepath.Dir(b)+"/../..", "cmd/operator/kodata")
--- a/test/resources/knativeresources.go
+++ b/test/resources/knativeresources.go
@@ -21,7 +21,14 @@ package resources
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
 	"time"
+
+	mf "github.com/manifestival/manifestival"
+	"knative.dev/operator/pkg/reconciler/common"
 
 	v1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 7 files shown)


test/resources/knativeresources.go
```diff
@@ -81,3 +88,38 @@ func IsKnativeDeploymentReady(dpList *v1.DeploymentList, expectedDeployments []s
...
+	for _, resource := range manifest.Filter(mf.ByKind("Deployment")).Resources() {
+		deployments = append(deployments, resource.GetName())
+	}
+	return removeDuplications(deployments)
+}
+
+// SetKodataDir will set the env var KO_DATA_PATH into the path of the kodata of this repository.
+func SetKodataDir() {
+	_, b, _, _ := runtime.Caller(0)
+	koPath := filepath.Join(filepath.Dir(b)+"/../..", "cmd/operator/kodata")
```
Cynocracy, line 111: ../.. is a bit of a magic string, is there a way we can get the top level dir without needing to know how nested we are?

test/upgrade/knativeeventing_preupgrade_test.go
```diff
@@ -40,14 +44,20 @@ func TestKnativeEventingPreUpgrade(t *testing.T) {
...
-		expectedDeployments := []string{"eventing-controller", "eventing-webhook", "imc-controller",
-			"imc-dispatcher", "broker-controller", "broker-filter", "broker-ingress", "mt-broker-controller"}
+		keventing, err := clients.KnativeEventing().Get(names.KnativeEventing, metav1.GetOptions{})
+		if err != nil {
+			t.Fatalf("Failed to get KnativeEventing CR: %v", err)
+		}
+		resources.SetKodataDir()
+		defer os.Unsetenv(common.KoEnvKey)
+		// Based on the status.version, get the deployment resources.
+		expectedDeployments := resources.GetExpectedDeployments(t, keventing.Status.Version, "knative-eventing")
```
Cynocracy, line 60: One concern I have is that if the kodata dir were mistakenly emptied, we would be checking that there were *no* deployments, and we might not notice from this test. Can you add a check to ensure that there aren't 0 expected?

test/resources/knativeresources.go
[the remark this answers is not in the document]
houshengbo: I have not found a better way.

PR #131, opened 2020-06-08
Refactor transform logic to common and restore as a "stage"

jcrossley3:
We're using the trick of manifest.Append to create a deep-copy of `r.config`, in anticipation of creating the manifests dynamically according to spec.version.

Either way, the transform function can keep its signature to act as a stage, re-assigning the passed-by-reference manifest to a transformed one.


pkg/reconciler/common/transformers.go
```diff
@@ -32,3 +33,27 @@ func Transformers(ctx context.Context, obj v1alpha1.KComponent) []mf.Transformer
...
 		ResourceRequirementsTransform(obj.GetSpec().GetResources(), logger)}
 }
+
+// Transform will replace the passed-by-reference manifest with one
+// transformed by platform, common, and any extra passed in
+// TODO: deal with platforms better, i.e. encapsulate their clientset
+func Transform(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent, client kubernetes.Interface, extra ...mf.Transformer) error {
+	logger := logging.FromContext(ctx)
+	logger.Debug("Transforming manifest")
+	platforms := GetPlatforms(ctx)
```
markusthoemmes, line 43: Is this really present on the "reconciling" context? Where is it added?

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -113,25 +107,15 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *servingv1alpha1.Knat
 	return nil
 }
 
-// Transform the resources
-func (r *Reconciler) transform(ctx context.Context, instance *servingv1alpha1.KnativeServing) (mf.Manifest, error) {
+// transform re-assigns the passed manifest to one with common and
```
markusthoemmes, line 110:
```suggestion
// transform mutates the passed manifest to one with common and
```

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -74,8 +72,8 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *servingv1alpha1
 		}
 	}
 
-	manifest, err := r.transform(ctx, original)
-	if err != nil {
+	manifest := r.config.Append()
```
markusthoemmes, line 75: Might want to comment that this is a `Clone()` operation essentially.

pkg/reconciler/common/transformers.go
jcrossley3: Apparently not. :) Restored, thanks.

pkg/reconciler/knativeserving/knativeserving.go
jcrossley3: Done

pkg/reconciler/knativeserving/knativeserving.go
jcrossley3: Done

pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -107,11 +111,11 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ke *eventingv1alpha1.Kna
 	return nil
 }
 
-// transform re-assigns the passed manifest to one with common and
+// transform mutates the passed manifest to one with common and
 // platform transforms, plus any extras passed in
 func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, instance *eventingv1alpha1.KnativeEventing) error {
 	logger := logging.FromContext(ctx)
-	return common.Transform(ctx, manifest, instance, r.kubeClientSet,
+	return common.Transform(common.WithPlatforms(ctx, r.platform), manifest, instance, r.kubeClientSet,
```
markusthoemmes, line 118: If we add this only here anyway, shall we just pass it? That at least makes it typesafe and we pass a ton of parameters already anyway.

jcrossley3: Good point. I went a step further, making platform more abstract until we know better how it's going to be used.

jcrossley3:
My bad! I meant to squash those last two commits, and I accidentally squashed all of them.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -92,18 +93,15 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *servingv1alpha1.Knat
 
 	logger.Infow("Reconciling KnativeServing", "status", ks.Status)
 	stages := []func(context.Context, *mf.Manifest, *servingv1alpha1.KnativeServing) error{
+		r.transform,
```
houshengbo, line 96: Line 79 has called r.transform once, why do you call it another time in stages?

markusthoemmes: The other call is in the Finalize function, that's not part of the reconcile loop and doesn't go via this statement.

PR #132, opened 2020-06-10
Create the func RetrieveManifest to generate the manifest

houshengbo:
Fixes #

## Proposed Changes

* This PR creates a common function RetrieveManifest to generate the manifest, based on version, component and the mf.Client.

CC @jcrossley3 @markusthoemmes @evankanderson @Cynocracy 
This is fundamental.
Next, I plan to dynamically delete the obsolete resources, by comparing the target and current manifests.

### DIFF
--- a/pkg/reconciler/common/read_releases.go
+++ b/pkg/reconciler/common/read_releases.go
@@ -17,12 +17,17 @@ limitations under the License.
 package common
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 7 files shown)


jcrossley3:
> @jcrossley3 would you mind opening your PR?

Sure thing: https://github.com/knative/operator/pull/133

> If we use a path as the key for the cache, what path will it be, if we download the yamls online?

Both URL's and filepaths can be treated as string keys to the cache.

PR #135, opened 2020-06-11
Add Spec.Version to the API, but not the CRD yet

jcrossley3:
This should be a strict refactoring. Since it's not in the CRD, the
operator will always use the latest embedded version.

The main benefit here is a single, logical place to maintain the
mapping of API type (serving or eventing) to a specific directory
beneath kodata.


pkg/reconciler/common/releases.go
```diff
@@ -0,0 +1,140 @@
...
+
+const (
+	KoEnvKey = "KO_DATA_PATH"
+)
+
+var cache = map[string]mf.Manifest{}
+
+func TargetVersion(instance v1alpha1.KComponent) string {
+	switch {
+	case instance.GetSpec().GetVersion() != "":
```
Cynocracy, line 40: Can this ever be nil? I don't know too much about the semantics of how we parse that struct.

pkg/reconciler/common/releases.go
```diff
@@ -0,0 +1,140 @@
...
+	default:
+		return latestRelease(instance)
+	}
+}
+
+func TargetManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
+	return fetch(manifestPath(TargetVersion(instance), instance))
+}
+
+func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
```
Cynocracy, line 53: Can you add comments to InstalledManifest and TargetManifest? I think they're similar enough that it might be confusing to some.

pkg/reconciler/common/releases.go
```diff
@@ -0,0 +1,140 @@
...
+	return releaseTags, nil
+}
+
+// latestRelease returns the latest release tag available under kodata directory for Knative component.
+func latestRelease(instance v1alpha1.KComponent) string {
+	vers, err := allReleases(instance)
+	if err != nil {
+		panic(err)
+	}
+	// The versions are in a descending order, so the first one will be the latest version.
```
Cynocracy, line 138: I don't think this is true, semvers sorted lexographically sometimes fails.

For example:
1.1.0
1.10.0
1.2.0

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -45,10 +45,8 @@ type Reconciler struct {
 	kubeClientSet kubernetes.Interface
 	// operatorClientSet allows us to configure operator objects
 	operatorClientSet clientset.Interface
-	// config is the manifest of KnativeServing
-	config mf.Manifest
-	// targetVersion is the version of the KnativeServing manifest to install
-	targetVersion string
+	// manifest is empty but with a valid client and logger
```
Cynocracy, line 48: maybe Manifest is initialized empty?

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -74,12 +72,14 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *servingv1alpha1
 		}
 	}
 
-	// Appending nothing effectively results in a deep-copy clone
-	manifest := r.config.Append()
+	manifest, err := common.InstalledManifest(original)
+	if err != nil {
+		return err
+	}
+	manifest = r.manifest.Append(manifest)
```
Cynocracy, line 79: Reading this I'm starting to get a sense of the distinction between these two Manifest fetching methods.

So we initialize with the existing install, then apply the target install to it?

pkg/reconciler/common/releases.go
houshengbo: Yep, when the users do not specify anything in spec.version.

pkg/reconciler/common/releases.go
houshengbo: That's why we have implement a `sort.Slice` function in `allReleases`, making sure that the result of the list is in the descending order for semantic versions. The semver lib can compare two versions.

pkg/reconciler/knativeserving/knativeserving.go
houshengbo: The target manifest is what to install.
The installed manifest is the existing manifest, which has been installed. 
Since we about to support multiple versions of knative with one version of operator, installed manifest is what we have, and target manifest will be where we are going, in terms of upgrading and downgrading.

pkg/reconciler/knativeserving/knativeserving.go
houshengbo: There is a tricky point that knativeserving/knativeeventing reconciler and controller do not share the same context, but we need the correct context in controller to create the correct `client and logger`.
Initially, we thought of creating manifest in reconciler, but it seemed we have to leverage the context in controller to create client and logger.
We are only able to know which version to be installed(target) in reconciler, by spec.version.
As you see, manifest is prepared in controller, but filled in resources in reconciler.

pkg/reconciler/common/releases.go
```diff
@@ -0,0 +1,140 @@
...
+	KoEnvKey = "KO_DATA_PATH"
+)
+
+var cache = map[string]mf.Manifest{}
+
+func TargetVersion(instance v1alpha1.KComponent) string {
+	switch {
+	case instance.GetSpec().GetVersion() != "":
+		return instance.GetSpec().GetVersion()
+	case instance.GetStatus().GetVersion() != "":
```
houshengbo, line 42: @jcrossley3 @markusthoemmes 
Again, I personally lean on to this approach. If spec.version is nil, go to check status.version. 
We discussed multiple time in slack channel. It seemed both of you are for directly looking up latestRelease, if spec.version is nil.

pkg/reconciler/common/releases.go
Cynocracy: Aha! thanks, that makes sense. In that case overall this LGTM

pkg/reconciler/knativeserving/knativeserving.go
Cynocracy: I mostly meant // manifest is empty 
implied to me that it stays empty, whereas really it just *starts* empty.

pkg/reconciler/knativeserving/knativeserving.go
Cynocracy: Sounds reasonable, I still think it may be helpful to comment that here.

Cynocracy:
Overall /lgtm though there are still some minor points under discussion.

pkg/reconciler/knativeserving/knativeserving.go
jcrossley3: It does actually stay empty. Each time a reconcile appends to it, that creates a brand new, deep-copy of it -- recall all manifests are immutable.

pkg/reconciler/common/releases.go
jcrossley3: I agree we need to deal with this, but I was thinking this PR wouldn't care since `spec.Version` isn't yet in the CRD. Until it is, both `TargetManifest` and `InstalledManifest` will return the `latestRelease`. But that will still be a problem for any existing CR's with a `status.Version` unknown to this operator. I'm not sure how `FinalizeKind` handles errors, but I'd expect that situation to put us in an infinite loop. Maybe if `InstalledManifest` returns an error, we use `TargetManifest` instead and log a warning that some cluster-scoped resources might not get cleaned up?

pkg/reconciler/knativeserving/knativeserving.go
Cynocracy: Ah! Then the comment is useful :) I did find that a bit surprising.

It might be better to be a bit more explicit and construct the Manifest below, but I don't feel too strongly about it so long as the comment stays. Maybe we could add the word "immutable" here as well?

pkg/reconciler/common/releases.go
houshengbo: We can return a warning message for now.

houshengbo:
@jcrossley3 Would you mind adding some comments for some functions to better explain?

pkg/reconciler/common/releases.go
jcrossley3: Done

pkg/reconciler/knativeserving/knativeserving.go
jcrossley3: Done

pkg/reconciler/knativeserving/knativeserving.go
jcrossley3: Done

jcrossley3:
> @jcrossley3 Would you mind adding some comments for some functions to better explain?

Done

pkg/reconciler/common/releases.go
jcrossley3: I've added the above logic to `FinalizeKind` in both serving and eventing -- there's opportunity to factor them into a single common function, but I want to get some consensus on the correctness of the logic for the purposes of this PR. Thoughts?

houshengbo: This PR added the cache with the path as key, the logic to retrieve the installed and target manifests, added spec.version into the objects...
It looks in the correct direction to me. Further refactoring can be done based on this PR.

jcrossley3:
The most recent commit could arguably be its own PR. I can go either way.

pkg/reconciler/common/releases.go
markusthoemmes: I agree, this is a change of behavior and not what we've discussed extensively I think. We wanted to always return `latestRelease` if `spec.Version` isn't set.

pkg/reconciler/common/releases.go
```diff
@@ -0,0 +1,164 @@
...
+const (
+	KoEnvKey = "KO_DATA_PATH"
+)
+
+var cache = map[string]mf.Manifest{}
+
+// TargetVersion returns the version of the manifest to be installed
+// per the spec in the component. If spec.version is empty, but
+// status.version is set, that will be returned, since the expectation
+// is that the current installation should be reconciled. If both
```
markusthoemmes, line 43: I don't think that's the expectation today, it may be at some point in the future but isn't what we've discussed.

pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -45,10 +45,11 @@ type Reconciler struct {
...
 	operatorClientSet clientset.Interface
-	// config is the manifest of KnativeEventing
-	config mf.Manifest
-	// targetVersion is the version of the KnativeEventing manifest to install
-	targetVersion string
+	// manifest is empty, but with a valid client and logger. all
+	// manifests are immutable, and any created during reconcile are
+	// expected to be appended to this one, obviating the passing of
+	// client & logger
+	manifest mf.Manifest
```
markusthoemmes, line 52: I don't think I like the semantics here. It feels backwards to create an empty manifest and then append to it and imo it's a sign of the manifestival API being wrong in this case.

Maybe they need a few tweaks? I feel like the client and the logger should only be passed to the side-effecting functions, like `Apply`, `Delete` etc and not be passed around with the manifest itself, which is (as you say) just an immutable collection of resources.

pkg/reconciler/common/releases.go
```diff
@@ -0,0 +1,164 @@
...
+		return instance.GetSpec().GetVersion()
+	case instance.GetStatus().GetVersion() != "":
+		return instance.GetStatus().GetVersion()
+	default:
+		return latestRelease(instance)
+	}
+}
+
+// TargetManifest returns the manifest for the TargetVersion
+func TargetManifest(ctx context.Context, instance v1alpha1.KComponent, stages ...ReconcileStage) (mf.Manifest, error) {
```
markusthoemmes, line 58: Why are the reconcile stages part of this call? Don't they do sideeffects? Looking below it seems like you're using this to apply transforms to the manifests, maybe we should more clearly signify that. We don't want "real" reconcile operations on this function (and on `InstalledManifest` for that matter) I believe. That might purely be a naming thing.

pkg/apis/operator/v1alpha1/common.go
```diff
@@ -99,6 +104,10 @@ type CommonSpec struct {
 	// Override containers' resource requirements
 	// +optional
 	Resources []ResourceRequirementsOverride `json:"resources,omitempty"`
+
+	// Override containers' resource requirements
+	// +optional
+	Version string `json:"version,omitempty"`
```
markusthoemmes, line 110: Doesn't this make it part of the CRD effectively? Or are we dropping unknown keys via the OpenAPI schema?

pkg/reconciler/common/releases.go
jcrossley3: I can do that. I had it in my head that returning `latestRelease` would be a surprise to the user in light of the operator having the multi-version behavior.

pkg/reconciler/common/releases.go
jcrossley3: Ok

pkg/reconciler/knativeeventing/knativeeventing.go
jcrossley3: Awkward, maybe. Wrong seems harsh. I suppose I could include client/logger as options to Apply/Delete/etc.

Personally, I liked having a single manifest member instead of both client and logger members. Having any members at all presents challenges as we introduce shared logic between the two reconcilers. That's what the `Parent` function addressed. I suppose the context parameter can be used for that, too.

pkg/reconciler/common/releases.go
jcrossley3: I might've gotten a little carried away there, but I don't hate it. It was mainly to clean up the calling code, since I was getting the manifest, appending it to the root, and transforming it, not to mention checking for errors between each. And that was just in `FinalizeKind`, but in *both* the eventing and serving reconcilers. It was a lot of redundant code, and there'll be even more, because...

...we'll need to do the exact same thing in `deleteObsoleteResources` for the installed manifest when we compute the diffs to know what resources to delete.

Plus, we'll likely need to include the `ingress` stage in both places as well, and whatever future stages we create to construct the appropriate manifest per the CR spec. 

It just seemed logical to include optional "post processing" stages as parameters to the fetcher. And while I was at it, I realized I could re-use the logic to clean up the `ReconcileKind` function. It seemed like a win to me.

pkg/apis/operator/v1alpha1/common.go
jcrossley3: You're referring to the member or the json tag? I thought if it wasn't explicitly defined in the schema of the CRD yaml, it wasn't a part of the CRD. I don't know, tbh.

pkg/reconciler/common/releases.go
markusthoemmes: Yeah I think it's a naming thing for me. We shouldn't actively reconcile (aka read + write from the API for example) in these functions, right?

jcrossley3: Why not? It's the only reason we're fetching the manifest at all. :) 

I'm always open to better names, of course.

markusthoemmes: Well my confusion is that I wouldn't expect sideeffects to happen when I call a function named `TargetManifest` that returns a manifest. I do buy applying transformers to it straight away though. "Reconcile" is just overloaded in my head with a fixup-feedback-loop, which I wouldn't expect this to do.

pkg/apis/operator/v1alpha1/common.go
markusthoemmes: Depends on how the CRDs openAPI schema is specified. If it doesn't have a "preserve-unknown-keys" directive then yes, keys we don't specify there will just be dropped (which is what we want in this case).

pkg/reconciler/common/releases.go
jcrossley3: Done

pkg/reconciler/common/releases.go
jcrossley3: Reverted

pkg/apis/operator/v1alpha1/common.go
jcrossley3: What are you suggesting I do exactly?

houshengbo:
/lgtm
This PR cuts a milestone, by adding the necessary changes to the models, with changing the operator API, creating the functions to retrieve the target and installed manifests, based on spec.version or status.version, storing a manifest with valid client and logger for further usage to access all resources, etc.

We can accept this PR as a key transition.

For further manifestival API development, we can either change the manifestival lib or open separate PRs.

For ReconcileStages, we can open separate PRs.

How to handle `unknown status.version` in a certain version of operator can be tracked by the issue: https://github.com/knative/operator/issues/136.

PR #137, opened 2020-06-16
Add the logic to determine if upgrade or downgrade is eligible

### LINKED ISSUE #139: Operator should only support upgrade or downgrade to the adjacent MINOR version
**Describe the bug**
Upgrade and downgrade paths are limited to the next and the previous minor version.
Knative operator does not guarantee upgrade and downgrade support across multiple minor versions. 

**Expected behavior**
Reconcile function should check whether the target and installed versions are within one MINOR version.
If not, reconcile function logs an error message, informing the users that `operator only supports upgrade or downgrade to the adjacent minor version`.

**To Reproduce**
Steps to reproduce the behavior.

**Knative release version**
Master.

houshengbo:
Fixes #139

## Proposed Changes

* This PR added the logic to determine if upgrade or downgrade is eligible. If the target version is within +-1 version of the installed version, the logic function returns true.

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/operator/v1alpha1/common.go
+++ b/pkg/apis/operator/v1alpha1/common.go
@@ -33,6 +33,9 @@ const (
 	// DeploymentsAvailable is a Condition indicating whether or not the Deployments of
 	// the respective component have come up successfully.
 	DeploymentsAvailable apis.ConditionType = "DeploymentsAvailable"
+	// VersionMigrationEligible is a Condition indicating whether or not the current version of
+	// Knative component is eligible to upgrade or downgrade to the specified version.
+	VersionMigrationEligible apis.ConditionType = "VersionMigrationEligible"
 )
 
 // KComponent is a common interface for accessing meta, spec and status of all known types.
@@ -71,6 +74,12 @@ type KComponentStatus interface {
 	// MarkDeploymentsNotReady marks the DeploymentsAvailable status as false and calls out
 	// it's waiting for deployments.
 	MarkDeploymentsNotReady()
+
+	// MarkVersionMigrationEligible marks the VersionMigrationEligible status as true.
+	MarkVersionMigrationEligible()
+	// MarkVersionMigrationNotEligible marks the VersionMigrationEligible status as false with
+	// the given message.
+	MarkVersionMigrationNotEligible(msg string)
 
 	// MarkDependenciesInstalled marks the DependenciesInstalled status as true.
 	MarkDependenciesInstalled()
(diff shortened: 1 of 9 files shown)


jcrossley3:
Would you mind recounting whatever slack convo led us to this in an issue, please? If I was a user, I wouldn't want this. I would hope it's only temporary until upstream fully embraces the operator.

pkg/reconciler/common/releases.go
```diff
@@ -63,6 +65,52 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
...
+	currentMajor := semver.Major(current)
+	targetMajor := semver.Major(target)
+	if currentMajor != targetMajor {
+		// All the official releases of Knative are under the same Major version number. If target and current versions
+		// are different in terms of major version, upgrade or downgrade is not supported.
+		// TODO We need to deal with the the case of bumping major version later.
+		return false
+	}
+
+	var err error
```
evankanderson, line 88: nit: I don't think you need to declare `err` here given the `:=` on the next line.

pkg/reconciler/common/releases.go
```diff
@@ -63,6 +65,52 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
...
+	currentMinor, err := strconv.Atoi(strings.TrimPrefix(semver.MajorMinor(current), currentMajor+"."))
+	if err != nil {
+		return false
+	}
+
+	targetMinor, err := strconv.Atoi(strings.TrimPrefix(semver.MajorMinor(target), targetMajor+"."))
```
evankanderson, line 94: Another nit, I'd find the following easier to read:
```suggestion
	targetMinor, err := strconv.Atoi(strings.Split(semver.MajorMinor(target), ".")[1])
```

Or even:

```suggestion
	targetMinor, err := strconv.Atoi(strings.Split(target, ".")[1])
```

houshengbo:
@evankanderson I added a Condition called VersionMigrationEligible to indicate whether the spec.version is able to migrate to.

pkg/apis/operator/v1alpha1/common.go
```diff
@@ -33,6 +33,9 @@ const (
 	// DeploymentsAvailable is a Condition indicating whether or not the Deployments of
 	// the respective component have come up successfully.
 	DeploymentsAvailable apis.ConditionType = "DeploymentsAvailable"
+	// VersionMigrationEligible is a Condition indicating whether or not the current version of
+	// Knative component is eligible to upgrade or downgrade to the specified version.
+	VersionMigrationEligible apis.ConditionType = "VersionMigrationEligible"
```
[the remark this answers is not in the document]
jcrossley3: +1 for `VersionCompatible`. Too late to change?

PR #138, opened 2020-06-16
Introduce Stages abstraction to simplify manifest post-processing

jcrossley3:
This enables all of the stages except the transform functions (which
are the only ones requiring downcasting of the KComponents) to be
moved out of the reconcilers into a common package.

It'll also be handy when obsolete resources are calculated from
diff'ing the target and installed manifests as all of the stages must
be applied to both.

Similarly, the ingress determination should be added to the stages in
FinalizeKind and (the forthcoming) deleteObsoleteResources functions,
along with any other stages that might potentially append to the
constructed manifests.

### DIFF
--- a/pkg/reconciler/common/stages.go
+++ b/pkg/reconciler/common/stages.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+
+	mf "github.com/manifestival/manifestival"
+	"knative.dev/operator/pkg/apis/operator/v1alpha1"
+	"knative.dev/pkg/logging"
+)
+
+type Stage func(context.Context, *mf.Manifest, v1alpha1.KComponent) error
+type Stages []Stage
+
+func (stages Stages) Execute(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent) error {
+	for _, stage := range stages {
+		if err := stage(ctx, manifest, instance); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func TargetStage(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent) error {
+	m, err := TargetManifest(instance)
+	if err != nil {
+		return err
+	}
+	*manifest = manifest.Append(m)
+	return nil
+}
+
+func InstalledOrTargetStage(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent) error {
+	logger := logging.FromContext(ctx)
+	m, err := InstalledManifest(instance)
+	if err != nil {
+		// TODO: return the oldest instead of the latest?
+		logger.Error("Unable to fetch installed manifest, trying target", err)
+		m, err = TargetManifest(instance)
+	}
+	if err != nil {
+		return err
+	}
+	*manifest = manifest.Append(m)
+	return nil
+}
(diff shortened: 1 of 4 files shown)


pkg/reconciler/common/stages.go
```diff
@@ -0,0 +1,70 @@
...
+		if err := stage(ctx, manifest, instance); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// AppendTarget mutates the passed manifest by appending one
+// appropriate for the passed KComponent
+func AppendTarget(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent) error {
```
markusthoemmes, line 45: Looking at the code where this is used, I'm wondering if we'd be better of adding transformers in here too, i.e. by doing this:

```golang
func AppendTargetManifest(transformers []mf.Transformer, platform Extension) Stage {
  return func(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent) {
    // Do your loading code
    // Also apply the transformers
  }
}
```

That'd get rid of the needed typecasts as the `r.transform` function completely vanishes and it makes getting a manifest a one-step process.

---

Thinking further, why don't we add all the necessary shenanigans to  the `TargetManifest` function? I.e. make it:

```golang
func TargetManifest(base mf.Manifest, instance KComponent, platform Extension, transformers mf.Transformer...) (mf.Manifest, error)
```

My reasoning behind the second approach is that I find the current pointer overriding a bit backwards too. We define a manifest variable before reconcilation, that we then pass as a pointer **into** reconcilation and rely on the first stages to override that pointer with something useful. Instead, the proposed `TargetManifest` function would return the correct manifest, with all mutations applied and we can just feed that into the reconcile stages statically. Does that make sense?

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -75,57 +75,45 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *servingv1alpha1
...
-		manifest, err = common.TargetManifest(original)
-	}
-	if err != nil {
-		logger.Error("Unable to fetch target manifest, no cluster-scoped resources will be finalized", err)
+	logger.Info("Deleting cluster-scoped resources")
+	// Create new, empty manifest with valid client and logger
+	manifest := r.manifest.Append()
+
+	// TODO: add ingress, etc
+	stages := common.Stages{common.AppendInstalled, r.transform}
```
markusthoemmes, line 83: I find this one strange in particular. I don't think of these as reconcilation stages at all. The suggestion above would clear that out though.

pkg/reconciler/common/stages.go
jcrossley3: Consider the introduction of the `ingress` behavior. That's not a transformation, it's a filter (or an append, depending on how we do it). And `ingress` won't be the only part of what I'm attempting to model here as a pipeline of sorts that munges the fetched manifest according to the CR spec.

pkg/reconciler/knativeserving/knativeserving.go
jcrossley3: I don't think of them as reconciliation stages either. I specifically removed the word "reconcile" out of the name to avoid this confusion. But they are stages in the construction of the manifest, and I expect there will be more, e.g. `ingress`

pkg/reconciler/common/stages.go
houshengbo: @markusthoemmes @jcrossley3 
I am thinking of the next step of comparing the target and installed manifests.
I lean on to having the function `AppendTarget`, returning the expected target manifest, WITHOUT `platform Extension, transformers mf.Transformer`, because we will UNINSTALL the diffs in installed manifest but not in target manifest. I think of the steps as finding out these diffs at the beginning from `the RAW manifests`, transforming them with necessary transformers, and removing them.

By RAW manifests, I mean original manifest contents with correct `client and logger`.

If you add params `platform Extension, transformers mf.Transformer`, we are kind of forcing the 
usage of `platform Extension, transformers mf.Transformer` to retrieve the target manifest. I hope we can have a func available to return the RAW manifest only.

markusthoemmes: My proposals above are butchered by the ingress being filters indeed. Please ignore.

@houshengbo Is there a situation where you need the **entire** raw manifest without the transformers etc. applied? If so: Where and why?

houshengbo: The only benefit to have the RAW manifests and compare them is to find out the delta between target and installed manifests first, then apply the transformers to the delta for further deletion, as a refactoring of the current `deleteObsoleteResources`.

I thought this approach was more performant than transforming all resources, and finding out the delta.

If we do not care the performance from this aspect, I would say we go with the more logical approach, by transforming manifests first, then compare.

PR #145, opened 2020-06-18
Add the knative 0.15.1 releases into kodata

houshengbo:
## Proposed Changes

* This PR adds the manifests of 0.15.1 into kodata.

### DIFF
--- a/cmd/operator/kodata/knative-serving/0.15.1/1-serving-crds.yaml
+++ b/cmd/operator/kodata/knative-serving/0.15.1/1-serving-crds.yaml
@@ -1,0 +1,602 @@
+# Copyright 2019 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


matzew:
@houshengbo  will there be a 0.15.x release of the opeartor?  If so, will the `release 0.15` branch be updated ? :thinking:

houshengbo:
yes. this will be backported into release 0.15.

cmd/operator/kodata/knative-serving/0.15.1/1-serving-crds.yaml
```diff
@@ -0,0 +1,602 @@
+# Copyright 2019 The Knative Authors
```
Cynocracy, line 1: Should we just always apply the CRDs early in the Operator logic?

cmd/operator/kodata/knative-serving/0.15.1/5-serving-storage-version-migration.yaml
```diff
@@ -0,0 +1,44 @@
+# Copyright 2020 The Knative Authors
```
Cynocracy, line 1: What's the difference between this and 3-serving-storage-version-migration?

houshengbo: I do not see there is a `3-serving-storage-version-migration`.

cmd/operator/kodata/knative-serving/0.15.1/1-serving-crds.yaml
houshengbo: We can do that in the reconciler, but reconciler needs to look up the CRD by the label `knative.dev/crd-install: "true"`.

PR #147, opened 2020-06-19
Add the support to specify customized yamls

### LINKED ISSUE #227: Adding `spec.manifests` support to specify customized yamls
This is one feature documented in https://docs.google.com/document/d/1ehd5Yx4aSottuv7FYW4JeplNULJCs7nnDfy29y3NH8k/:

Operator CRs have supported `spec.version`. Next, we will offer admins the flexibility to specify links of customized yamls.

```
spec:
  version: 0.14.0
  manifests:
  - URL: https://github.com/knative/serving/releases/${version}/serving-core.yaml
  - URL: https://github.com/knative/serving/releases/${version}/serving-cert-manager.yaml
  - URL: https://github.com/knative/net-istio/releases/${version}/net-istio.yaml
```

Admins can specify the the manifests by the full URL links, or links with ${version} parameter to be replaced by the spec.version.
For example, if one URL is `https://github.com/knative/serving/releases/download/v${version}/serving-core.yaml`, operator will translate the link into `https://github.com/knative/serving/releases/download/v0.14.0/serving-core.yaml`.

With spec.manifests, admins can install any official release of knative, available online, but not available inside the operator, or any customized manifests available online.

houshengbo:
Fixes #227

## Proposed Changes

* Adds the logic to download manifests online, if they are unavailable locally in kodata.


houshengbo:
@jcrossley3 I added the logic to download the manifests.

pkg/reconciler/common/releases.go
```diff
@@ -23,17 +23,30 @@ import (
...
 
-var cache = map[string]mf.Manifest{}
+var (
+	cache = map[string]mf.Manifest{}
+
+	// ServingYamlNames is the string array containing all the artifacts to download for Knative serving
+	ServingYamlNames = []string{"serving-crds.yaml", "serving-core.yaml", "serving-hpa.yaml", "serving-storage-version-migration.yaml"}
+
+	// EventingYamlNames is the string array containing all the artifacts to download for Knative eventing
+	EventingYamlNames = []string{"eventing.yaml", "upgrade-to-%s.yaml", "storage-version-migration-%s.yaml"}
```
jcrossley3, line 48: I don't think we can assume these files will be a part of every version we might fetch. I think we need to determine these dynamically instead of hard-coding them.

houshengbo: @jcrossley3 How can we dynamically know them?

jcrossley3: I don't know. It may require some upstream operand coordination. But if the answer is "we can't" then I don't think we can claim to be able to reliably install a future operand release without cutting a corresponding operator release.

houshengbo: @jcrossley3 The path is based on the template `https://github.com/knative/%s/releases/download/v%s/%s`. The first %s is the component and the second is version number.
EventingYamlNames and ServingYamlNames save the file names, which are consistent for all releases(at least for now). As long as Knative still release the same files, we are fine to read them dynamically for each release.

jcrossley3: So when say, "serving-storage-version-migration.yaml", is removed from a release, what then? Cut a new release of the operator? You'll still have to deal with the fact that some versions supported by the operator contain that file and some don't.

markusthoemmes: I agree I think we need a naming scheme that we agree with upstream.

I'd expect `serving-crds` and `serving-core` to be stable. The jobs might need a prefix or something so we can list the artifacts through the releases API and fetch all of the upgrade jobs necessary. Or we agree with the upstream's that there's always a YAML called `serving-upgrade.yaml`, that contains all necessary upgrade resources.

I agree we can't hardcode those as is.

houshengbo: https://github.com/knative/serving/issues/8436
https://github.com/knative/eventing/issues/3386
I logged two issues for serving and eventing, to track this request.

pkg/reconciler/common/releases.go
```diff
@@ -85,9 +98,50 @@ func componentDir(instance v1alpha1.KComponent) string {
...
+func componentURL(version string, instance v1alpha1.KComponent) string {
+	component := "serving"
+	urlList := []string{}
+	switch instance.(type) {
+	case *v1alpha1.KnativeServing:
+		urlList = ServingYamlNames
+	case *v1alpha1.KnativeEventing:
+		component = "eventing"
+		for _, value := range EventingYamlNames {
+			if strings.Contains(value, "%s") {
```
markusthoemmes, line 110: This seems obsolete, there's no such resource above.

pkg/reconciler/common/releases.go
```diff
@@ -85,9 +98,50 @@ func componentDir(instance v1alpha1.KComponent) string {
...
+	// Create the comma-separated string as the URL to retrieve the manifest
+	if len(urlList) == 0 {
+		return ""
+	}
+
+	result := fmt.Sprintf(URLLinkTemplate, component,
+		version, urlList[0])
+
+	for i := 1; i < len(urlList); i++ {
+		result = fmt.Sprintf("%s,"+URLLinkTemplate, result, component,
```
markusthoemmes, line 127: This will leave a dangling `,`.

Please use `strings.Join` to join an array of strings with a delimiter in this case.

pkg/reconciler/common/releases.go
```diff
@@ -85,9 +98,50 @@ func componentDir(instance v1alpha1.KComponent) string {
...
+			version, urlList[i])
```
markusthoemmes, line 128: This entire thing could be:

```golang
urls := make([]string, 0, len(urlList)
for _, file := range urlList {
  urls = append(urls, fmt.Sprintf(URLLinkTemplate, component, version, file)
}
return strings.Join(urls, ",")
```

houshengbo: Done.

pkg/reconciler/common/releases.go
houshengbo: Done.

pkg/reconciler/common/releases.go
houshengbo: Done.

pkg/reconciler/common/releases.go
```diff
@@ -85,9 +98,42 @@ func componentDir(instance v1alpha1.KComponent) string {
...
+	component := "serving"
+	urlList := []string{}
+	switch instance.(type) {
+	case *v1alpha1.KnativeServing:
+		urlList = ServingYamlNames
+	case *v1alpha1.KnativeEventing:
+		component = "eventing"
+		for _, value := range EventingYamlNames {
+			urlList = append(urlList, value)
+		}
```
markusthoemmes, line 111: Why not just the same line as above? `urlList = EventingYamlNames`

houshengbo: Oops, EventingYamlNames changed.
Thx.

pkg/reconciler/common/releases.go
```diff
@@ -85,9 +98,40 @@ func componentDir(instance v1alpha1.KComponent) string {
 	return ""
 }
 
+func componentURL(version string, instance v1alpha1.KComponent) string {
+	component := "serving"
+	urlList := []string{}
```
markusthoemmes, line 103:
```suggestion
	var urlList []string
```

No need to allocate an object here. I'd also rename this. It doesn't contain a list of URLs but of filenames.

pkg/apis/operator/v1alpha1/common.go
```diff
@@ -95,6 +97,11 @@ type KComponentStatus interface {
 	// SetVersion sets the currently installed version of the component.
 	SetVersion(version string)
 
+	// GetManifests gets the url links of the manifests
+	GetManifests() string
+	// SetManifests sets the url links of the manifests
```
Cynocracy, line 102: What's the format?

pkg/apis/operator/v1alpha1/common.go
```diff
@@ -59,6 +59,8 @@ type KComponentSpec interface {
 	GetResources() []ResourceRequirementsOverride
 	// GetVersion gets the version to be installed
 	GetVersion() string
+	// GetManifests gets the list of manifests to be installed
```
Cynocracy, line 62: Is this redundant with GetVersion? Can they ever disagree?

pkg/reconciler/common/releases_test.go
```diff
@@ -25,8 +25,20 @@ import (
 	util "knative.dev/operator/pkg/reconciler/common/testing"
 )
 
+const (
+	VERSION              = "0.16.0"
+	SERVING_CORE         = "https://github.com/knative/serving/releases/download/v" + VERSION + "/serving-core.yaml"
```
Cynocracy, line 30: This means our tests depend on github to run.

Is there any way to fake out the actual HTTP call to keep this test hermetic?

houshengbo: No. https links are just example.
URL can be any valid links locally or remotely.

pkg/apis/operator/v1alpha1/common.go
evankanderson: I was just going to ask how/whether version interacts with this field. :grin:

pkg/reconciler/common/releases_test.go
evankanderson: Maybe we can make this a relative `file://` URL to a file that we create in a temp dir?

evankanderson:
Last question, which is truly an open one:

* Do we want to land this before or after the release cut?

pkg/apis/operator/v1alpha1/common.go
houshengbo: Good question:
single url: just the full url link or url link with ${version}
multiple urls: concatenate them together with comma as delimiter.

houshengbo:
First, we need to count on the admins to specify the spec.version matching the manifests. About picking up the version from the manifests, it is not in this PR.

Second, if manifest url links are invalid, then operator fails to get the manifests, CR is set to error state.(error message is logged. I do not think we put the error message in the CR so far)

Third, URL links can be set as the full links without any version.
Also, if the URL consists of the version number equal to the spec.version, we are able to specify it as `https://github.com/knative/serving/releases/v${version}/serving-core.yaml` as example. ${version} will be replaced with spec.version.

pkg/reconciler/common/releases_test.go
houshengbo: we can definitely do that. as long as the path is accessible, we can use it as the link.

pkg/apis/operator/v1alpha1/common.go
houshengbo: The manifests in status is set automatically by the operator. Users do not care it.

Cynocracy: Interesting. So getManifests is a way for the Operator to expose what versions it 'knows about'?

pkg/reconciler/common/releases_test.go
Cynocracy: +1, I think that makes it clearer that the test doesn't depend on github.

pkg/apis/operator/v1alpha1/common.go
Cynocracy, line 62: Is "to be installed" valid? Maybe "which should ultimately be installed" to keep it a bit more declarative?

Cynocracy: Any update here?

pkg/reconciler/common/releases_test.go
```diff
@@ -43,10 +55,53 @@ func TestRetrieveManifestPath(t *testing.T) {
...
 		version:   "0.14.2",
 		expected:  koPath + "/knative-eventing/0.14.2",
+	}, {
+		name: "Valid Knative Serving URLs",
+		component: &v1alpha1.KnativeServing{
+			Spec: v1alpha1.KnativeServingSpec{
+				CommonSpec: v1alpha1.CommonSpec{
+					Manifests: []v1alpha1.Manifest{{
+						Url: SERVING_CORE,
+					}, v1alpha1.Manifest{
```
mattmoor, line 65: Format Go code:
```suggestion
					}, {
```

pkg/reconciler/common/releases_test.go
```diff
@@ -43,10 +55,53 @@ func TestRetrieveManifestPath(t *testing.T) {
...
+		version:  VERSION,
+		expected: SERVING_CORE + "," + SERVING_HPA,
+	}, {
+		name: "Valid Knative Eventing URLs",
+		component: &v1alpha1.KnativeEventing{
+			Spec: v1alpha1.KnativeEventingSpec{
+				CommonSpec: v1alpha1.CommonSpec{
+					Manifests: []v1alpha1.Manifest{{
+						Url: EVENTING_CORE,
+					}, v1alpha1.Manifest{
```
mattmoor, line 80: Format Go code:
```suggestion
					}, {
```

pkg/reconciler/common/releases_test.go
```diff
@@ -43,10 +55,53 @@ func TestRetrieveManifestPath(t *testing.T) {
...
+		version:  VERSION,
+		expected: EVENTING_CORE + "," + IN_MEMORY_CHANNEL,
+	}, {
+		name: "Valid Knative Serving URLs with the version parameter",
+		component: &v1alpha1.KnativeServing{
+			Spec: v1alpha1.KnativeServingSpec{
+				CommonSpec: v1alpha1.CommonSpec{
+					Manifests: []v1alpha1.Manifest{{
+						Url: SERVING_VERSION_CORE,
+					}, v1alpha1.Manifest{
```
mattmoor, line 95: Format Go code:
```suggestion
					}, {
```

pkg/apis/operator/v1alpha1/common.go
evankanderson: I think I'd feel more comfortable about this interface as `[]string` rather than "`string` with internal comma separation". It seems easier for computers to parse, too.

houshengbo: This one under KComponentStatus just indicate what the actual paths are for manifests.
If the manifests consist of multiple links, which can be local paths, can be remote paths, etc, they will be concatenated together, with the delimiter `,`.

pkg/apis/operator/v1alpha1/common.go
```diff
@@ -173,3 +189,9 @@ type ResourceRequirementsOverride struct {
 	// The desired ResourceRequirements
 	corev1.ResourceRequirements
 }
+
+// Manifest enables the user to specify the links to the manifests' URLs
+type Manifest struct {
+	// The link of the manifest URL
+	Url string `json:"URL"`
```
evankanderson, line 196: Is this an [`apis.URL`](https://github.com/knative/pkg/blob/master/apis/url.go#L30)?

I think it's not, because it has some form of string replacement supported, right? Should we document the replacement language?

pkg/reconciler/common/releases.go
```diff
@@ -132,9 +136,49 @@ func componentDir(instance v1alpha1.KComponent) string {
 	return ""
 }
 
+func componentURL(version string, instance v1alpha1.KComponent) string {
+	manifests := instance.GetSpec().GetManifests()
+	// Create the comma-separated string as the URL to retrieve the manifest
+	if len(manifests) == 0 {
+		return ""
+	}
```
evankanderson, line 144: Do we need this early-exit code anymore?

pkg/reconciler/common/releases.go
```diff
@@ -32,10 +32,14 @@ import (
...
 
 const (
+	// KoEnvKey is the key of the environment variable to specify the path to the ko data directory
 	KoEnvKey = "KO_DATA_PATH"
 )
 
-var cache = map[string]mf.Manifest{}
+var (
+	cache           = map[string]mf.Manifest{}
+	versionVariable = "${version}"
```
evankanderson, line 41: Can this be `const`?

pkg/reconciler/common/releases.go
```diff
@@ -132,9 +136,49 @@ func componentDir(instance v1alpha1.KComponent) string {
...
+func componentURL(version string, instance v1alpha1.KComponent) string {
+	manifests := instance.GetSpec().GetManifests()
+	// Create the comma-separated string as the URL to retrieve the manifest
+	if len(manifests) == 0 {
+		return ""
+	}
+
+	urls := make([]string, 0, len(manifests))
+	for _, manifest := range manifests {
+		url := strings.ReplaceAll(manifest.Url, versionVariable, version)
```
evankanderson, line 148: Would using [`os.Expand`](https://golang.org/pkg/os/#Expand) with the following function be easier to extend in the future:

```
func varExpander(vars map[string]string) func(string) string {
  return func(s string) {
    return vars[s]
  }
}
```

(Obviously, this could be a later refactor, so it's not a priority.)

pkg/apis/operator/v1alpha1/common.go
houshengbo: That's correct. We should document the string replacement usage.

pkg/reconciler/common/releases.go
houshengbo: we can remove it

pkg/reconciler/common/releases.go
houshengbo: Yep, we can change it later.

houshengbo:
@evankanderson The cache needs to be rethought as well.
If the links are exactly the same, but the remote manifests change the contents, it seems operator can not refresh the contents.
I am not sure whether we should allow cleaning-up the cache, or enforcing to load from the paths.

pkg/apis/operator/v1alpha1/common.go
Cynocracy: Ah, ok. So in the status, the Version shown will always be the selected one.

Do we still need a GetVersion in the spec, though?

Cynocracy:
Vincent and I agreed that allowing unset version when it's not needed by the manifest string is a good path forward, but that he would do that in a separate PR.

/lgtm

PR #149, opened 2020-06-19
Comparing target & installed manifests to delete obsolete resources

### LINKED ISSUE #148: Improve `deleteObsoleteResources` by  deleting the resources Not In the target manifest
**Describe the bug**
The func `deleteObsoleteResources` is rigid, because it maintains all the resources to delete.
As manifestival is able to compare two manifests, we are able to find out the diffs between two manifests and delete the diffs.

**Expected behavior**
Delete the resources, which exist in the previous manifest, but do not exist in the target manifest to install.

**Knative release version**
Master

jcrossley3:
Fixes #148 but leaves some opportunity for refactoring the functions
in each controller to a shared package. I wasn't happy with any of my
efforts to deal with the "parent" manifests (with client/logger) and
build stages (e.g. ingress, transform), though.

### DIFF
--- a/pkg/reconciler/knativeeventing/knativeeventing.go
+++ b/pkg/reconciler/knativeeventing/knativeeventing.go
@@ -22,7 +22,6 @@ import (
 
 	mf "github.com/manifestival/manifestival"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	clientset "knative.dev/operator/pkg/client/clientset/versioned"
@@ -76,13 +75,12 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1alpha1.Knativ
 	}
 
 	logger.Info("Deleting cluster-scoped resources")
-	manifest := r.manifest.Append()
-	stages := common.Stages{common.AppendInstalled, r.transform}
-	if err := stages.Execute(ctx, &manifest, original); err != nil {
+	manifest, err := r.installed(ctx, original)
+	if err != nil {
 		logger.Error("Unable to fetch installed manifest; no cluster-scoped resources will be finalized", err)
 		return nil
 	}
-	return common.Uninstall(&manifest)
+	return common.Uninstall(manifest)
 }
 
 // ReconcileKind compares the actual state with the desired, and attempts to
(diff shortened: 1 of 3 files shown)


pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -99,7 +97,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ke *v1alpha1.KnativeEven
 		r.ensureFinalizerRemoval,
 		r.install,
 		r.checkDeployments,
-		r.deleteObsoleteResources,
+		r.deleteObsoleteResources(ctx, ke),
```
houshengbo, line 100: What is the benefit of keeping deleteObsoleteResources as a stage?
If we only runs `installed.Filter(mf.None(mf.In(*manifest))).Delete()`, after the Stages, what is bad about that?

jcrossley3: You could, as long as you create the installed manifest before you execute the stages. I thought it was cleaner to put that logic in the `deleteObsoleteResources` function instead of `ReconcileKind`.

houshengbo: What about this?
Change this line into `r.deleteObsoleteResources,`.
Change the function `deleteObsoleteResources` into 
```
func (r *Reconciler) deleteObsoleteResources(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent) error {
   ...
   ...
   return installed.Filter(mf.None(mf.In(*manifest))).Delete()
}
```

jcrossley3: In your example, where is `installed` coming from. You can't create it "inside the pipeline" because one of the previous stages might set `instance.Status.Version` to something other than what it was when it was originally passed to `ReconcileKind`.

houshengbo: installed, err := r.installed(ctx, instance)
I omit all the lines above.

jcrossley3: If you do that, you'll never delete any obsolete resources. The version of your installed manifest will be identical to your target manifest because `r.install` updates `status.version` upon successful completion: https://github.com/knative/operator/blob/9a134e09fe836907592f8d2cafb4f2b491eb493b/pkg/reconciler/common/install.go#L50

Because both the installed version *and* the target version are both determined by looking at the CR instance, you must calculate both using the _same_ instance _before_ any stages are executed that might mutate the instance.

You could argue that calculating the installed version in `ReconcileKind` is more clear and obviates the `NoOp` stages, but I liked having just the stages in there, and traded a bit of complexity for that.

Before we go that route, I'll see if I can add an e2e test to this PR that demonstrates the condition.

houshengbo: Got it. Thx.

houshengbo:
/lgtm
@jcrossley3's comments resolved my concerns.
As this PR is a milestone as transition from status to dymanic reconciling mechanism to install target manifest, and deleting the obsolete delta, we accept it.
Further implementation will move on.

PR #152, opened 2020-06-23
Refactor image transformation

trshafer:
This allows for an ImageTransformer that is not based on v1alpha1.Registry

## Proposed Changes

* Internal refactor that exposes an internal image transformer

**Release Note**
NONE

```release-note

```

### DIFF
--- a/pkg/reconciler/common/images.go
+++ b/pkg/reconciler/common/images.go
@@ -41,34 +41,80 @@ var (
 	delimiter             = "/"
 )
 
+// ImageTransformer is an interface for transforming images passed to the ResourceImageTransformer
+type ImageTransformer interface {
+	HandleImage(containerName, parent string, useDefault bool) (string, bool)
+	HandleImagePullSecrets(imagePullSecrets []corev1.LocalObjectReference, log *zap.SugaredLogger) []corev1.LocalObjectReference
+}
+
+// registryImageTransformer is a v1alpha1.Registry specific transformer
+type registryImageTransformer struct {
+	registry *v1alpha1.Registry
+}
+
+var _ ImageTransformer = (*registryImageTransformer)(nil)
+
+func (rit *registryImageTransformer) HandleImage(containerName, parent string, useDefault bool) (string, bool) {
+	if image, ok := rit.registry.Override[parent+delimiter+containerName]; ok {
+		return image, true
+	}
+	if image, ok := rit.registry.Override[containerName]; ok {
+		return image, true
+	}
+	if !useDefault {
+		return "", false
+	}
+	return replaceName(rit.registry.Default, containerName), true
+}
+
+func replaceName(imageTemplate string, name string) string {
+	return strings.ReplaceAll(imageTemplate, containerNameVariable, name)
+}
+
+func (rit *registryImageTransformer) HandleImagePullSecrets(imagePullSecrets []corev1.LocalObjectReference, log *zap.SugaredLogger) []corev1.LocalObjectReference {
+	if len(rit.registry.ImagePullSecrets) > 0 {
+		log.Debugf("Adding ImagePullSecrets: %v", rit.registry.ImagePullSecrets)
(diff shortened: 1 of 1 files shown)


pkg/reconciler/common/images.go
```diff
@@ -41,34 +41,80 @@ var (
 	delimiter             = "/"
 )
 
+// ImageTransformer is an interface for transforming images passed to the ResourceImageTransformer
```
Cynocracy, line 44: Maybe we could upstream this into manifestival?

trshafer: That would work for me. @jcrossley3 what do you think about Manifestival providing some out of the box transformations. The specific `v1alpha1.Registry` transformation would continue to live in this operator code but the `ImageTransformer` would live in `manifestival`. I'm happy to do the port.

jcrossley3: I like having MF provide ootb transformers, e.g. namespace and ownerref, but is there much value to other non-knative MF users without a default impl?

Cynocracy: IMO, there would be value to providing Identity and ChangeRepo implementations (and likely more, if we thought about it :). I think needing to swap out a personal repository while testing / use different repositories for different locations/shards/environments would be fairly frequent use cases for anything manipulating a manifest.

pkg/reconciler/common/images.go
```diff
@@ -41,34 +41,80 @@ var (
...
+type ImageTransformer interface {
+	HandleImage(containerName, parent string, useDefault bool) (string, bool)
+	HandleImagePullSecrets(imagePullSecrets []corev1.LocalObjectReference, log *zap.SugaredLogger) []corev1.LocalObjectReference
```
Cynocracy, line 47: Handle -> Transform?

Handle makes me think it's a 'visit' function more than something returning another value.

Alternatively: a doc would work

pkg/reconciler/common/images.go
```diff
@@ -41,34 +41,80 @@ var (
...
 // ImageTransform updates image with a new registry and tag
 func ImageTransform(registry *v1alpha1.Registry, log *zap.SugaredLogger) mf.Transformer {
+	rit := &registryImageTransformer{
+		registry: registry,
+	}
+	return ResourceImageTransformer(rit, log)
+}
+
+// ResourceImageTransformer takes an ImageTransformer and transform images across resources
+func ResourceImageTransformer(imageTransformer ImageTransformer, log *zap.SugaredLogger) mf.Transformer {
```
Cynocracy, line 91: why public?

Cynocracy:
ah, thought that would LGTM as well, guess not.

/lgtm

Cynocracy:
Ah, I did not know it would immediately merge :/

pkg/reconciler/common/images.go
trshafer: Making this public is the point of the change. It allows for other people to make an `ImageTransformer` and pass it to `ResourceImageTransformer`.

pkg/reconciler/common/images.go
trshafer: We could provide a base impl for swapping a regex with another string:

```
type RegexpReplacer struct {
  ToReplace string
  ReplaceWith string
}

var _ ImageTransformer = (*RegexpReplacer)(nil)

func (rr *RegexpReplacer) HandleImage(containerName, parent string, useDefault bool) (string, bool) {
  strings.ReplaceAll(containerName, rr.ToReplace, rr.ReplaceWith)
}

func replaceName(imageTemplate string, name string) string {
	return strings.ReplaceAll(imageTemplate, containerNameVariable, name)
}

func (rit *registryImageTransformer) HandleImagePullSecrets(imagePullSecrets []corev1.LocalObjectReference, log *zap.SugaredLogger) []corev1.LocalObjectReference {
	return imagePullSecrets
}
```

trshafer:
@Cynocracy here's a revert PR: https://github.com/knative/operator/pull/153. but if this PR looks good to others then we can just keep it and close the revert pr.

PR #164, opened 2020-06-25
Refactor image replacement interface for explicit support of each typ…

trshafer:
…e of object

## Proposed Changes

* Internal refactor of the exposed ImageTransformer. This allows for greater flexibility of the implementor.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/common/images.go
+++ b/pkg/reconciler/common/images.go
@@ -43,7 +43,9 @@ var (
 
 // ImageTransformer is an interface for transforming images passed to the ResourceImageTransformer
 type ImageTransformer interface {
-	HandleImage(containerName, parent string, useDefault bool) (string, bool)
+	ImageForContainer(container *corev1.Container, parentName string) (string, bool)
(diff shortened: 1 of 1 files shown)


pkg/reconciler/common/images.go
```diff
@@ -43,7 +43,9 @@ var (
 
 // ImageTransformer is an interface for transforming images passed to the ResourceImageTransformer
 type ImageTransformer interface {
-	HandleImage(containerName, parent string, useDefault bool) (string, bool)
+	ImageForContainer(container *corev1.Container, parentName string) (string, bool)
```
Cynocracy, line 46: Should/could we have a default impl that only requires you to populate this method (for callers who don't have different behavior between env vars, cached images, and containers)?

Cynocracy: Spoke offline, the main issue is the use of env var name as container names (which may be a bug?)

Can you add a TODO to investigate that?

trshafer: Fixed. Changed to resource name

PR #167, opened 2020-06-30
Run e2e tests in arbitrary namespace

### LINKED ISSUE #166: e2e tests should run in an arbitrary namespace
While using the operator to install into a namespace other than `knative-serving`, I discovered [a manifestival bug](https://github.com/manifestival/manifestival/issues/55). The bug wasn't revealed until I actually tried to create a Knative Service, so our current e2e tests may need to be enhanced to replicate that, but regardless, I think we should be running our tests inside an arbitrary namespace just to attempt to catch things like this (I believe the conversion webhooks in the CRD are a recent addition).

jcrossley3:
Fixes #166

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -12,11 +12,10 @@ Before running the integration, please make sure you have installed
 install custom resource for operator or knative-serving installed in your
 cluster.
 
-Ensure required namespaces exist:
+Ensure required testing namespace exists:
 
 ```bash
-kubectl create namespace knative-serving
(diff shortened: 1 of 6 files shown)


jcrossley3:
@houshengbo can you identify exactly why the update test is failing? I've never been able to run it locally as I think it assumes GKE. Some of the infra stuff may assume the `knative-serving` namespace, too. I'm hoping we can PR changes upstream to make that configurable, if so.

PR #170, opened 2020-07-01
Change test cases to use the same namespace for serving and eventing

houshengbo:
Fixes #

## Proposed Changes

* As we are about to run both knative serving and eventing under the same namespace, test cases verification needs to be changed.
* This PR dropped the verification for the number of deployments, but verify whether the expected deployments are available.
* Knative serving ns is set by the env var SYSTEM_NAMESPACE, and Knative eventing ns is set by the env var TEST_EVENTING_NAMESPACE in shell scripts.

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -18,17 +18,17 @@
 source $(dirname $0)/../vendor/knative.dev/test-infra/scripts/e2e-tests.sh
 
 # The previous operator release.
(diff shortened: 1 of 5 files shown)


markusthoemmes:
Please add some context on what the plan is here. PR title + description don't tell me anything :thinking:

houshengbo:
0.14 and 0.15 both pick up the latest manifestival to resolve the issue with non-default ns conversion webhook.
Need to wait for new patch releases for 14 and 15.
After that, test scripts need to update the previous version to install.

PR #178, opened 2020-07-08
[WIP] Fix a case where the upgrade job was stuck if its image was changed a bit late in the spec override section

aliok:
Fixes #

## Proposed Changes

* Fix a case where the upgrade job was stuck if its image was changed a bit late in the spec override section
*
*

**Release Note**

```release-note

```

### DIFF
--- a/pkg/reconciler/knativeeventing/knativeeventing.go
+++ b/pkg/reconciler/knativeeventing/knativeeventing.go
@@ -23,6 +23,7 @@ import (
 	mf "github.com/manifestival/manifestival"
 	"go.uber.org/zap"
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


jcrossley3:
@aliok As discussed in slack, I think this is accounting for a situation that will only occur due to our downstream architecture, so let's handle it there.

pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -123,6 +125,7 @@ func (r *Reconciler) reconcile(ctx context.Context, ke *eventingv1alpha1.Knative
 		r.ensureFinalizer,
 		r.initStatus,
 		r.install,
+		r.checkJobs,
```
houshengbo, line 128: I actually would like to see a new transformer to convert the Job named `storage-version-migration`.
@jcrossley3 what do u think?

houshengbo:
https://github.com/knative/operator/pull/180
I opened a PR to rename job name. I do not think I am doing the same work as Ali. Correct?

aliok:
@houshengbo 
Thanks for the heads up.
It is a different problem I think.
I was testing something in this PR to fix a problem downstream at OpenShift side.

For the curious, it was about where our downstream stuff was too late to change the image overrides in the spec and thus the upstream operator was complaining that it cannot change the image of the job.
We decided to tackle this problem downstream though.

PR #180, opened 2020-07-08
Add the tranformer for job resources

houshengbo:
## Proposed Changes

* This PR added tranformer to the job named `storage-version-migration` in eventing.


pkg/reconciler/knativeeventing/common/job.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil
+	}
+}
+
+func updateJobNameLabel(job *batchv1.Job) error {
+	// Change the job name
+	jobName := job.GetName()
+	if strings.HasPrefix(jobName, StorageVersionMigration) && !strings.HasPrefix(jobName,
+		StorageVersionMigrationEventing) {
+		suffix := jobName[len(StorageVersionMigration):len(jobName)]
```
mattmoor, line 75: Format Go code:
```suggestion
		suffix := jobName[len(StorageVersionMigration):]
```

pkg/reconciler/knativeeventing/common/job.go
```diff
@@ -0,0 +1,95 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	// StorageVersionMigration is the prefix of the job name to be updated
+	StorageVersionMigration = "storage-version-migration"
+
+	// StorageVersionMigrationEventing is the label name that the job STORAGE_VERSION_MIGRATION will change into.
+	StorageVersionMigrationEventing = "storage-version-migration-eventing"
+)
```
jcrossley3, line 37: I think we need this logic for all jobs, not just these.

pkg/reconciler/knativeeventing/common/job.go
```diff
@@ -0,0 +1,95 @@
...
+		labels["app"] = StorageVersionMigrationEventing
+		job.SetLabels(labels)
+	}
+
+	// Change the labels in spec.template.metadata
+	labels = job.Spec.Template.GetLabels()
+	if labels["app"] == StorageVersionMigration {
+		labels["app"] = StorageVersionMigrationEventing
+		job.SetLabels(labels)
+	}
```
jcrossley3, line 92: I hadn't considered the labels, but I suppose they have to match the names? Is that true? I'm not sure how exactly they're used.

houshengbo: I just verified. We do not need to change the labels. We only need different job names.

pkg/reconciler/common/job.go
```diff
@@ -0,0 +1,91 @@
...
+		if strings.HasSuffix(jobName, component) {
+			jobName = fmt.Sprintf("%s-%s", jobName, version)
+		} else {
+			jobName = fmt.Sprintf("%s%s", jobName, suffix)
+		}
+		job.SetName(jobName)
+	}
+
+	return nil
+}
```
jcrossley3, line 91: This seems overly complex to me. The conversion isn't necessary because the unstructured type has all the metadata fields you need. And I don't think we need to worry about whether the component name is already in the name -- I don't think it currently is upstream -- we only care about ensuring the names are unique. Also, the convention for `generateName` values is that they will already include a dash '-' suffix, which the following suggestion accounts for:

```suggestion
func JobTransform(obj v1alpha1.KComponent) mf.Transformer {
	return func(u *unstructured.Unstructured) error {
		if u.GetKind() == "Job" {
			component := "serving"
			if _, ok := obj.(*v1alpha1.KnativeEventing); ok {
				component = "eventing"
			}
			if u.GetName() == "" {
				u.SetName(fmt.Sprintf("%s%s-%s", u.GetGenerateName(), component, TargetVersion(obj)))
			} else {
				u.SetName(fmt.Sprintf("%s-%s-%s", u.GetName(), component, TargetVersion(obj)))
			}
		}
		return nil
	}
}
```

pkg/reconciler/common/job_test.go
```diff
@@ -0,0 +1,111 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+
+	util "knative.dev/operator/pkg/reconciler/common/testing"
+)
+
+const (
+	StorageVersionMigration         = "storage-version-migration"
+	StorageVersionMigrationEventing = "storage-version-migration-eventing"
+	StorageVersionMigrationServing  = "storage-version-migration-serving"
```
jcrossley3, line 35: I don't think we need to care about these, so I would omit them. Who's to say where the component name might show up in the name? I don't think we need to worry about that.

pkg/reconciler/common/job_test.go
```diff
@@ -0,0 +1,111 @@
...
+	}, {
+		name: "ChangeNameWithEventingForEventingJob",
+		component: &v1alpha1.KnativeEventing{
+			Spec: v1alpha1.KnativeEventingSpec{
+				CommonSpec: v1alpha1.CommonSpec{
+					Version: "0.16.0",
+				},
+			}},
+		job:      createJob(StorageVersionMigrationEventing),
+		expected: StorageVersionMigrationEventing + "-0.16.0",
```
jcrossley3, line 84: Omit these. And replace with a test for a Job with generateName set.

pkg/reconciler/common/job_test.go
```diff
@@ -0,0 +1,111 @@
...
+	return batchv1.Job{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Job",
+			APIVersion: "batch/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+	}
+}
```
jcrossley3, line 111: Add a `gen` param here and adjust your above tests to call both `createJob(StorageVersionMigration, "")` as well as `createJob("", StorageVersionMigration)`.

```suggestion
func createJob(name, gen string) batchv1.Job {
	return batchv1.Job{
		TypeMeta: metav1.TypeMeta{
			Kind:       "Job",
			APIVersion: "batch/v1",
		},
		ObjectMeta: metav1.ObjectMeta{
			Name: name,
			GenerateName: gen + "-",
		},
	}
}
```

pkg/reconciler/common/transformers.go
```diff
@@ -30,7 +30,9 @@ func transformers(ctx context.Context, obj v1alpha1.KComponent) []mf.Transformer
 		mf.InjectNamespace(obj.GetNamespace()),
 		ImageTransform(obj.GetSpec().GetRegistry(), logger),
 		ConfigMapTransform(obj.GetSpec().GetConfig(), logger),
-		ResourceRequirementsTransform(obj.GetSpec().GetResources(), logger)}
+		ResourceRequirementsTransform(obj.GetSpec().GetResources(), logger),
+		JobTransform(obj, logger),
```
jcrossley3, line 34: I don't think we need the logger for anything here (probably not in the others either, but we can deal with that later). Whatever error is caught by `stages.Execute` will be logged.

PR #181, opened 2020-07-08
Add 0.16.0 manifests of knative

houshengbo:
## Proposed Changes

* This PR added the released 0.16.0 knative.

### DIFF
--- a/test/resources/knativeresources.go
+++ b/test/resources/knativeresources.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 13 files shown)


test/resources/knativeresources.go
```diff
@@ -177,10 +169,21 @@ func IsKnativeObsoleteResourceGone(clients *test.Clients, namespace string, obsR
 		gvr := apis.KindToResource(resource.GroupVersionKind())
 		var err error
 		if resource.GetNamespace() != "" {
-			// This is a namespaced resource
+			// Verify all namespaced resources, except jobs.
```
Cynocracy, line 172: Why not Jobs?

test/resources/knativeresources.go
```diff
@@ -177,10 +169,21 @@ func IsKnativeObsoleteResourceGone(clients *test.Clients, namespace string, obsR
...
+			}
 			_, err = clients.Dynamic.Resource(gvr).Namespace(namespace).Get(resource.GetName(), metav1.GetOptions{})
 		} else {
-			// This is a clustered resource
+			// TODO(#1): If APIVersion is the only different field between two resources with
+			// one being v1 and the other being v1beta1, the dynamic client can access both of
+			// them in the cluster. Before we find out the reason, we skip verifying CRDs and
+			// webhooks for all clustered resources.
+			switch strings.ToLower(resource.GetKind()) {
+			case "customresourcedefinition", "validatingwebhookconfiguration", "mutatingwebhookconfiguration":
```
Cynocracy, line 184: I'd be more OK with this if we're confident we're going to get rid of this for job upgrades later.

Are we definitely moving forward with that? Can we add a TODO to clean this up?

test/resources/knativeresources.go
houshengbo: We added a job transformation, so the job name won't match the one of the launched job. It is meaningless to check jobs.

Cynocracy: OK, so long as we don't create a new job every reconcile loop, I suppose.

I wouldn't say it's meaningless to check jobs, though. They may be necessary for the correctness of the upgrade.

PR #188, opened 2020-07-10
Skip the verification of CRD's and webhooks

### LINKED ISSUE #189: No need to verify CRD's and webhooks in upgrade test cases
If version is the only difference between two CRD's or webhooks, manifestival considers them as different, but they should still exist in the cluster after upgrade. 

Test cases for upgrade should not verify them

houshengbo:
Fixes #189

## Proposed Changes

* If version is the only difference between two CRD's or webhooks, manifestival considers them as different, but they should still exist in the cluster after upgrade. There is no need to verify them in the postupgrade tests.

### DIFF
--- a/test/resources/knativeresources.go
+++ b/test/resources/knativeresources.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 	"time"
 
@@ -180,7 +181,11 @@ func IsKnativeObsoleteResourceGone(clients *test.Clients, namespace string, obsR
 			// This is a namespaced resource
 			_, err = clients.Dynamic.Resource(gvr).Namespace(namespace).Get(resource.GetName(), metav1.GetOptions{})
 		} else {
-			// This is a clustered resource
+			// Verify all clustered resources, except CRDs and webhooks.
+			switch strings.ToLower(resource.GetKind()) {
+			case "customresourcedefinition", "validatingwebhookconfiguration", "mutatingwebhookconfiguration":
+				continue
+			}
 			_, err = clients.Dynamic.Resource(gvr).Get(resource.GetName(), metav1.GetOptions{})
 		}
 		if !apierrs.IsNotFound(err) {
--- a/test/resources/knativeresources.go
+++ b/test/resources/knativeresources.go
@@ -113,7 +113,7 @@ func IsKnativeDeploymentReady(dpList *v1.DeploymentList, expectedDeployments []s
 // GetExpectedDeployments will return an array of deployment resources based on the version for the knative
 // component.
 func GetExpectedDeployments(t *testing.T, instance v1alpha1.KComponent) (mf.Manifest, []string) {
-	manifest, err := common.InstalledManifest(instance)
+	manifest, err := common.TargetManifest(instance)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


test/resources/knativeresources.go
```diff
@@ -180,7 +181,11 @@ func IsKnativeObsoleteResourceGone(clients *test.Clients, namespace string, obsR
 			// This is a namespaced resource
 			_, err = clients.Dynamic.Resource(gvr).Namespace(namespace).Get(resource.GetName(), metav1.GetOptions{})
 		} else {
-			// This is a clustered resource
+			// Verify all clustered resources, except CRDs and webhooks.
+			switch strings.ToLower(resource.GetKind()) {
+			case "customresourcedefinition", "validatingwebhookconfiguration", "mutatingwebhookconfiguration":
+				continue
```
jcrossley3, line 187: Instead of ignoring these, shouldn't we verify they still exist? I'm confused by the inclusion of webhook configs here, too. I understand why we shouldn't delete CRD's but shouldn't the webhook configs be properly deleted?

houshengbo: @jcrossley3 Here is the difficulty: we get the delta, containing a validatingwebhookconfiguration. We do not know whether validatingwebhookconfiguration still exists.
If version is the only difference for validatingwebhookconfiguration, it should exist.
If name is different, it should not exist.

houshengbo: For validatingwebhookconfiguration, if we install v1beta1, then upgrade to v1, v1beta1 version can still be accessed.

houshengbo: This is the reason why I suggested to consider two resources the same, when `version` is the only difference in manifestival.

jcrossley3: I don't understand what you mean. I'd like for someone more knowledgeable about API evolution to confirm we're doing the right thing. Essentially, if i have a v1beta1 "Thing" installed and I then create a v1 "Thing" with the same name, should we delete the v1beta1 "Thing"?

/cc @pmorie @markusthoemmes

jcrossley3: > This is the reason why I suggested to consider two resources the same, when `version` is the only difference in manifestival.

Sure we could do that, but is it the right thing to do? If it is, then what's the point of having an `apiVersion` at all?

jcrossley3: Isn't that more a problem with the dynamic client than manifestival? Does some cache need to be refreshed?

markusthoemmes: IIRC, it should be possible to just override the v1beta1 thing with the new v1 thing **without** deleting it intermittently.

houshengbo: @jcrossley3 @markusthoemmes I added a TODO comment to find out the reason, for skipping the checking of CRD's and webhooks.
I am not sure if this is a bug with dynamic client, with kube v1 and v1beta1 versioning, or this is the correct behavior.

Cynocracy: > IIRC, it should be possible to just override the v1beta1 thing with the new v1 thing without deleting it intermittently.

That sounds right to me. I've been trying to wrap my head around GroupVersionKind/GroupKind etc recently, and boy does it get complicated.

PR #190, opened 2020-07-13
WIP: Add spec version into operator CR's and change the downgrade in scripts

houshengbo:
## Proposed Changes

* This PR added spec.version and change the implementation of downgrade in prow scripts.


test/resources/knativeserving.go
```diff
@@ -87,7 +87,10 @@ func WaitForConfigMap(name string, client *kubernetes.Clientset, fn func(map[str
 
 // IsKnativeServingReady will check the status conditions of the KnativeServing and return true if the KnativeServing is ready.
 func IsKnativeServingReady(s *v1alpha1.KnativeServing, err error) (bool, error) {
-	return s.Status.IsReady(), err
+	if s.Spec.Version == "" {
+		return s.Status.IsReady(), err
+	}
+	return s.Status.IsReady() && s.Spec.Version == s.Status.Version, err
```
markusthoemmes, line 93: Aren't we checking Generation vs. ObservedGen yet? If so, I think this is redundant.

PR #203, opened 2020-07-14
Prevent infinite reconcile loop

### LINKED ISSUE #185: Something removes Job's 'spec.ttlSecondsAfterFinished', causing an infinite loop in reconciler
Neither the `KnativeServing` nor the `KnativeEventing` CR's ever completely reconcile, because it appears *something* removes the `spec.ttlSecondsAfterFinished` field from the created Job and manifestival keeps trying to put it back.

At least, that's what I *think* is happening.

jcrossley3:
This fixes #185 but has nothing to do with the TTL for completed jobs, which turned out to be a red herring.

This should be backported to 0.15.x, I believe.

### DIFF
--- a/config/300-eventing.yaml
+++ b/config/300-eventing.yaml
@@ -137,6 +137,9 @@ spec:
           type: object
         status:
           properties:
+            observedGeneration:
+              description: The generation last processed by the controller
+              type: integer
             conditions:
               description: The latest available observations of a resource's current
                 state.
--- a/config/300-serving.yaml
+++ b/config/300-serving.yaml
@@ -174,6 +174,9 @@ spec:
         status:
           description: Status defines the observed state of KnativeServing
           properties:
+            observedGeneration:
+              description: The generation last processed by the controller
+              type: integer
             conditions:
               description: The latest available observations of a resource's current
                 state.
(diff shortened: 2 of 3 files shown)


pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -88,8 +88,10 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1alpha1.Knativ
 // converge the two.
 func (r *Reconciler) ReconcileKind(ctx context.Context, ks *v1alpha1.KnativeServing) pkgreconciler.Event {
 	logger := logging.FromContext(ctx)
-	ks.Status.InitializeConditions()
-	ks.Status.ObservedGeneration = ks.Generation
+	if len(ks.Status.Conditions) == 0 {
+		ks.Status.InitializeConditions()
+		ks.Status.ObservedGeneration = ks.Generation
```
markusthoemmes, line 93: We definitely want to run this on every reconcile as we want to make sure that `ObservedGeneration` is consistently updated.
Are we only churning through `resourceVersion` or also through `generation`? The latter would indicate we're constantly updating the Spec.

PR #207, opened 2020-07-20
Add spec version into operator CRD's 

houshengbo:
One version of operator bundles multiple available knative released versions.  Users can pick up one available version of knative to install.

## Proposed Changes

* This PRs added the filed `spec.version` into the operator CRD's, giving users the option to specify the official release to install with operator.

### DIFF
--- a/config/300-eventing.yaml
+++ b/config/300-eventing.yaml
@@ -59,6 +59,9 @@ spec:
         spec:
           description: Spec defines the desired state of KnativeEventing
           properties:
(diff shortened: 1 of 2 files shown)


evankanderson:
Since this is an **alpha** API, this will give us feedback on whether this is the right direction or not.

Given the current status/history of the operator, I think collecting that feedback is probably more important than certainty on forward API shape **for right now**.

/lgtm

PR #235, opened 2020-08-12
Add the postdowngrade tests, making sure the downgrade is complete

houshengbo:
## Proposed Changes

* This PR improves the test cases by adding and running postdowngrade tests, after downgrading to the previous operator version.
* Upgrade and downgrade can by done by leveraging spec.version set to different values.

### DIFF
--- a/test/downgrade/knative_postdowngrade_test.go
+++ b/test/downgrade/knative_postdowngrade_test.go
@@ -1,0 +1,127 @@
+// +build postdowngrade
+
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


test/downgrade/knative_postdowngrade_test.go
```diff
@@ -0,0 +1,127 @@
...
+		preManifest, err := common.TargetManifest(instance)
+		if err != nil {
+			t.Fatalf("Failed to get KnativeServing manifest: %v", err)
+		}
+		expectedDeployments := resources.GetExpectedDeployments(preManifest)
+		util.AssertEqual(t, len(expectedDeployments) > 0, true)
+		resources.AssertKnativeDeploymentStatus(t, clients, names.Namespace, test.OperatorFlags.PreviousServingVersion,
+			expectedDeployments)
+		resources.AssertKnativeObsoleteResource(t, clients, names.Namespace,
+			targetManifest.Filter(mf.Not(mf.In(preManifest))).Resources())
```
Cynocracy, line 125: Is there a way for us to make a 'fake' downgrade for tests? This relies on the downgrade being 'correct' which seems fragile depending on how the operand implements things.

Don't feel too strongly about it though, as I think this is a good qualification improvement.

/lgtm

PR #236, opened 2020-08-13
Adding istio ignore annotation transformer for jobs

### LINKED ISSUE #159: Add the job sidecar transformer to all the job
**Describe the bug**
The resources Job needs to be have the annotation `"sidecar.istio.io/inject": "false"`.

**Expected behavior**
Operator needs to have a transformer func to add this annotation.

**Knative release version**
All.

AceHack:
/lint

Fixes #159

**Release Note**
```release-note
Automatically adds `sidecar.istio.io/inject: false` to any jobs
```

### DIFF
--- a/pkg/reconciler/common/job.go
+++ b/pkg/reconciler/common/job.go
@@ -20,24 +20,50 @@ import (
 	"fmt"
 
 	mf "github.com/manifestival/manifestival"
+	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes/scheme"
 	"knative.dev/operator/pkg/apis/operator/v1alpha1"
 )
+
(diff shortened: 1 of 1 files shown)


AceHack:
@houshengbo @Cynocracy These failures seem related to the recently added post downgrade tests.
https://github.com/knative/operator/pull/235

The changes I've made here add `sidecar.istio.io/inject: false` if it doesn't already exist on jobs.  Since it already exists on jobs, this results in basically a no-op today.  So I don't see how it could have affected these tests.  Thanks and let me know any suggestions you might have.

houshengbo:
@AceHack It is true that the downgrade is not stable.
It is hard for me to hack, coz when I tried locally, it turned out fine.
However, in the prow job of upgrade, the downgrade part fails every a few builds.
Sometimes, it fails to remove some obsolete resources; sometimes, it fails to go back to the old version for the deployment resource...

houshengbo:
@Cynocracy I am asking myself again: if upgrade is going fine, why do I care the downgrade?

PR #239, opened 2020-08-13
WIP: Set status.version once all expected deployments are ready

houshengbo:
## Proposed Changes

* This PR fixed the unstable downgrade in prow.
* We need to relocate the line setting the status.version. It should be moved behind we make sure deployments are available. If not, then any error in `CheckDeployments` function will break the `DeleteObsoleteResources`, since any error before will lead to no call of `DeleteObsoleteResources`, but status.version is changed. In that situation, we will never get manifests compared(The status.version has been set to the new version in the previous reconciling loop. The old version is set to target version. They are always equal.), and never remove the diffs.

**Release Note**

```release-note

```

### DIFF
--- a/pkg/reconciler/common/deployments.go
+++ b/pkg/reconciler/common/deployments.go
@@ -50,6 +50,7 @@ func CheckDeployments(ctx context.Context, manifest *mf.Manifest, instance v1alp
 		}
 	}
 	status.MarkDeploymentsAvailable()
+	status.SetVersion(TargetVersion(instance))
 	return nil
 }
 
--- a/pkg/reconciler/common/install.go
+++ b/pkg/reconciler/common/install.go
@@ -52,7 +52,6 @@ func Install(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComp
 		return fmt.Errorf("failed to apply non rbac manifest: %w", err)
 	}
 	status.MarkInstallSucceeded()
-	status.SetVersion(TargetVersion(instance))
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


pkg/reconciler/common/deployments.go
```diff
@@ -50,6 +50,7 @@ func CheckDeployments(ctx context.Context, manifest *mf.Manifest, instance v1alp
 		}
 	}
 	status.MarkDeploymentsAvailable()
+	status.SetVersion(TargetVersion(instance))
```
jcrossley3, line 53: I would prefer we not do this here, for the same reason we shouldn't do it after the install: some temporary error in `DeleteObsoleteResources` might prevent a subsequent reconcile to delete the obsolete resources after the temporary error condition is resolved. Instead, I think we should set `status.Version` in the reconciler itself, as the very last stage.

pkg/reconciler/common/install.go
```diff
@@ -52,7 +52,6 @@ func Install(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComp
 		return fmt.Errorf("failed to apply non rbac manifest: %w", err)
 	}
 	status.MarkInstallSucceeded()
-	status.SetVersion(TargetVersion(instance))
```
jcrossley3, line 55: I understand why we can't do this here, but I still think it's important to denote somewhere in the CR exactly which version the operator installed. The current implementations of `MarkInstallSucceeded` call `MarkTrue` but let's have them call `MarkTrueWithReason` instead and include the `TargetVersion` in that reason.

PR #245, opened 2020-08-16
Add (simple) error reporting when operator stages are unable to install.

### LINKED ISSUE #216: Custom Object doesn't provide feedback on invalid `spec.version`
**Describe the bug**
If `spec.version` is an invalid value, the KnativeServing or KnativeEventing object stays in a state that appears to be unreconciled (`Ready=Unknown`) with no errors described

**Expected behavior**
An error condition should be set and `Ready` should be `False`.

**To Reproduce**
1. Install the 0.16 operator.

1. Create a KnativeServing object like so:
   ```yaml
   kind: KnativeServing
   metadata:
     name: knative-serving
     namespace: knative-serving
   spec:
     version: "dummy"
   ```

1. Once reconciled, the object looks like this:

   ```yaml
   kind: KnativeServing
   metadata:
     annotations:
       kubectl.kubernetes.io/last-applied-configuration: |
         {"apiVersion":"operator.knative.dev/v1alpha1","kind":"KnativeServing","metadata":{"annotations":{},"name":"knative-serving","namespace":"knative-serving"},"spec":{"version":"dummy"}}
     creationTimestamp: "2020-07-26T15:15:21Z"
     finalizers:
     - knativeservings.operator.knative.dev
     generation: 1
     name: knative-serving
     namespace: knative-serving
     resourceVersion: "10700"
     selfLink: /apis/operator.knative.dev/v1alpha1/namespaces/knative-serving/knativeservings/knative-serving
     uid: ca90e686-2ab6-49fe-add3-f05de2f4dfed
   spec:
     version: dummy
   status:
     conditions:
     - lastTransitionTime: "2020-07-26T15:15:21Z"
       status: Unknown
       type: DependenciesInstalled
     - lastTransitionTime: "2020-07-26T15:15:21Z"
       status: Unknown
       type: DeploymentsAvailable
     - lastTransitionTime: "2020-07-26T15:15:21Z"
       status: Unknown
       type: InstallSucceeded
     - lastTransitionTime: "2020-07-26T15:15:21Z"
       status: Unknown
       type: Ready
     - lastTransitionTime: "2020-07-26T15:15:21Z"
       status: "True"
       type: VersionMigrationEligible
     observedGeneration: 1
   ```

Note that all of the statuses (except `VersionMigrationEligible`) have status "Unknown"

**Knative release version**
0.16.0

**Additional context**
I'm not sure this needs a cherrypick, but we should address it for 0.17 for sure.

evankanderson:
Fixes #216 

## Proposed Changes

* In several places, we `return err` to indicate that the reconciliation can't make progress. In these places, we should also set the status.

**Release Note**

```release-note
Errors during the installation process are now reported in the custom object status.
```


evankanderson:
Sample output:

```yaml
kind: KnativeServing
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operator.knative.dev/v1alpha1","kind":"KnativeServing","metadata":{"annotations":{},"name":"knative-serving","namespace":"knative-serving"},"spec":{"version":"test"}}
  creationTimestamp: "2020-07-26T15:15:21Z"
  finalizers:
  - knativeservings.operator.knative.dev
  generation: 2
  name: knative-serving
  namespace: knative-serving
  resourceVersion: "223570"
  selfLink: /apis/operator.knative.dev/v1alpha1/namespaces/knative-serving/knativeservings/knative-serving
  uid: ca90e686-2ab6-49fe-add3-f05de2f4dfed
spec:
  version: test
status:
  conditions:
  - lastTransitionTime: "2020-07-26T15:15:21Z"
    status: Unknown
    type: DependenciesInstalled
  - lastTransitionTime: "2020-07-26T15:15:21Z"
    status: Unknown
    type: DeploymentsAvailable
  - lastTransitionTime: "2020-08-16T18:09:51Z"
    message: 'Install failed with message: stat : no such file or directory'
    reason: Error
    status: "False"
    type: InstallSucceeded
  - lastTransitionTime: "2020-08-16T18:09:51Z"
    message: 'Install failed with message: stat : no such file or directory'
    reason: Error
    status: "False"
    type: Ready
  - lastTransitionTime: "2020-07-26T15:15:21Z"
    status: "True"
    type: VersionMigrationEligible
  observedGeneration: 2
```

evankanderson:
It would probably be better if we also had the error coming back from the installation failure include the path that was trying to be `stat`ed, but that seems like a separate PR.

pkg/reconciler/common/stages.go
```diff
@@ -34,6 +34,7 @@ type Stages []Stage
 func (stages Stages) Execute(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent) error {
 	for _, stage := range stages {
 		if err := stage(ctx, manifest, instance); err != nil {
+			instance.GetStatus().MarkInstallFailed(err.Error())
```
jcrossley3, line 37: This could yield confusing behavior, I think. Each "stage" is expected to set the condition that's appropriate for it. So the install stage should be the only stage to ever set the `InstallSucceeded` condition, i.e. call `MarkInstallFailed`. The term "install" in this context means "applied the manifest without error", but we won't be `Ready` until all the other conditions are met, e.g. `DeploymentsAvailable`.

So the way you have this written, the install stage might call `MarkInstallSucceeded` but if the subsequent stage that checks deployments returns an error, this line will overwrite it. That's not giving the user an accurate record of what happened, IMO.

pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -100,6 +100,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ke *v1alpha1.KnativeEven
 	}
 
 	if err := r.extension.Reconcile(ctx, ke); err != nil {
+		ke.Status.MarkInstallFailed(err.Error())
```
jcrossley3, line 103: ibid

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -103,6 +103,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *v1alpha1.KnativeServ
 	}
 
 	if err := r.extension.Reconcile(ctx, ks); err != nil {
+		ks.Status.MarkInstallFailed(err.Error())
```
jcrossley3, line 106: ibid

pkg/reconciler/common/stages.go
evankanderson: Good point!

I've switched these to all simply set `Ready = False` if there's an error, with an "Internal error" message. (And tests added.)

I may see about upstreaming this into `pkg` as well, as it seems like any Reconcile that returns an error should probably signal that in the resource's condition somehow.

evankanderson:
Updated, and added a test for a stage that actually sets a failure condition and then returns an error as well.

pkg/reconciler/common/stages.go
```diff
@@ -96,3 +100,20 @@ func DeleteObsoleteResources(ctx context.Context, instance v1alpha1.KComponent,
...
 	}
 }
+
+// SetErrorConditionIfIndeterminate marks the instance's Conditions as not Ready
+// if no existing error has been signaled. If Ready is already False, it does
+// not change the existing error reporting. This exists to ensure that errors
+// are reported to the user even if the underlying functions do not explicitly
+// set conditions on the instance.
+func SetErrorConditionIfIndeterminate(err error, instance v1alpha1.KComponent) {
+	if !instance.GetStatus().GetCondition(apis.ConditionReady).IsFalse() {
```
jcrossley3, line 110: The double negative here is confusing. What does "not isFalse" mean exactly here? GetStatus().IsReady()?

pkg/reconciler/common/stages.go
```diff
@@ -96,3 +100,20 @@ func DeleteObsoleteResources(ctx context.Context, instance v1alpha1.KComponent,
...
+		instance.GetStatus().SetConditions(apis.Conditions{
+			apis.Condition{
+				Type:    apis.ConditionReady,
+				Status:  v1.ConditionFalse,
+				Reason:  "StageError",
+				Message: "Internal error: " + err.Error(),
+			}})
```
jcrossley3, line 117: Overwriting all the conditions with just a single one here makes me nervous. Is there no risk of losing any condition set by a previous stage?

jcrossley3 (commented):
I think I would rather account for what I think you're accounting for by adding another condition (internal error?) instead of overwriting all of them. 

Or am I completely misunderstanding this use case?

maximilien (approved):
New here... but seems fine to me
/ok-to-test

PR #257, opened 2020-08-21
Add the label operator.knative.dev/release into the operator resource for release

houshengbo:
Fixes #

## Proposed Changes

* Add the label `operator.knative.dev/release` into the operator resource.

**Release Note**

```release-note

```

### DIFF
--- a/hack/generate-yamls.sh
+++ b/hack/generate-yamls.sh
@@ -52,6 +52,12 @@ rm -fr ${YAML_OUTPUT_DIR}/*.yaml
 # Generated Knative Operator component YAML files
 readonly OPERATOR_YAML=${YAML_OUTPUT_DIR}/operator.yaml
 
+if [[ -n "${TAG}" ]]; then
+  LABEL_YAML_CMD=(sed -e "s|operator.knative.dev/release: devel|operator.knative.dev/release: \"${TAG}\"|")
+else
+  LABEL_YAML_CMD=(cat)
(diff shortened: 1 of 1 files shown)


hack/generate-yamls.sh
```diff
@@ -52,6 +52,12 @@ rm -fr ${YAML_OUTPUT_DIR}/*.yaml
 # Generated Knative Operator component YAML files
 readonly OPERATOR_YAML=${YAML_OUTPUT_DIR}/operator.yaml
 
+if [[ -n "${TAG}" ]]; then
+  LABEL_YAML_CMD=(sed -e "s|operator.knative.dev/release: devel|operator.knative.dev/release: \"${TAG}\"|")
```
Cynocracy, line 56: Don't you want $() here?

hack/generate-yamls.sh
```diff
@@ -52,6 +52,12 @@ rm -fr ${YAML_OUTPUT_DIR}/*.yaml
...
+else
+  LABEL_YAML_CMD=(cat)
```
Cynocracy, line 58: Is this to be ""? I find it a bit confusing, why this style?

hack/generate-yamls.sh
Cynocracy: Ah, I think I get it now. I'm not familiar with the () style for constructing strings. neat

PR #266, opened 2020-09-03
Skip the version checking for network ingress deployment

houshengbo:
Fixes #

## Proposed Changes

* The version of network ingress is not necessary matching the version of the knative serving. When we verify the deployment, we should skip the version checking on network ingress resource.

**Release Note**

```release-note

```


aliok (approved):
/lgtm
/approve

PR #291, opened 2020-10-05
Add 0.18.0 eventing and serving and remove 0.14.x

houshengbo:
## Proposed Changes

* This PR adds the 0.18.0 Knative manifests and removes 0.14.x manifests.
* As the istio yamls have been from serving to net-istio, this PR changes the repository to access the script to install istio.


evankanderson (commented):
/lgtm

Git came up with some funny moves, but this looks reasonable for now. I still want to figure out a way to generate these automatically.

Issue #297, opened 2020-10-06
TestContinuousEventsPropagationWithProber failing on `waiting for forwarder scales to zero: wathola-forwarder: timed out waiting for the condition`


houshengbo:
If we set E2E_UPGRADE_TESTS_SERVING_USE to true, eventing installed by operator will end up with the following test case failing:

```
 === FAIL: test/upgrade TestContinuousEventsPropagationWithProber (177.16s)
    test_runner.go:151: namespace is : "test-continuous-events-propagation-with-prober-lmddv"
2020-10-06T03:05:50.261Z	INFO	prober/prober.go:140	Using namespace for probe testing: test-continuous-events-propagation-with-prober-lmddv
    creation.go:294: Creating v1beta1 broker default
2020-10-06T03:05:50.394Z	INFO	prober/configuration.go:78	Deploying config map: "test-continuous-events-propagation-with-prober-lmddv/wathola-config"
2020-10-06T03:05:50.395Z	DEBUG	prober/configuration.go:61	Fetching default broker URL for ns test-continuous-events-propagation-with-prober-lmddv
2020-10-06T03:05:51.589Z	DEBUG	prober/configuration.go:72	default broker URL for ns test-continuous-events-propagation-with-prober-lmddv is http://broker-ingress.knative-eventing.svc.cluster.local/test-continuous-events-propagation-with-prober-lmddv/default
    creation.go:253: Creating configmap wathola-config
    creation.go:329: Creating v1beta1 trigger wathola-trigger-step
    creation.go:329: Creating v1beta1 trigger wathola-trigger-finished
2020-10-06T03:05:51.886Z	INFO	prober/receiver.go:42	Deploy of receiver pod: wathola-receiver
2020-10-06T03:05:52.001Z	INFO	prober/receiver.go:94	Deploy of receiver service: wathola-receiver
2020-10-06T03:05:52.259Z	DEBUG	prober/receiver.go:129	Node port for service: wathola-receiver is 30152
2020-10-06T03:05:52.259Z	INFO	prober/forwarder.go:36	Deploy forwarder knative service: wathola-forwarder
    tracker.go:153: Waiting for all KResources to become ready
2020-10-06T03:06:44.370Z	INFO	prober/sender.go:32	Deploy sender pod: wathola-sender
2020-10-06T03:06:44.478Z	INFO	lib/await.go:44	Wait for receiver be ready: wathola-receiver
[...]

PR #299, opened 2020-10-06
Don't wait for Ksvc to scale to zero

cardil:
Relates to #297

## Proposed Changes

* Skips waiting for forwarder ksvc component to scale to zero before routing traffic.

**Release Note**

```release-note
NONE
```


houshengbo:
/lgtm
/approve
This PR will tentatively resolve the failure of the prow upgrade test. We will dig into the root cause of the issue `failing to scale to zero`, and resolve it with separate PRs.

PR #305, opened 2020-10-14
lint: previous condition includes return statement, pop else statement

lberk:
## Proposed Changes

* The else statement isn't needed, the previous if statement includes a non-optional return
*
*


houshengbo (commented):
/lgtm
/approve
Thx for the contribution!

PR #307, opened 2020-10-16
Spelling (with checksums)

antoineco:
@jsoref I'm only opening this PR to show you what's missing in #301 for the tests to pass (see last 2 commits).

image

There is a tool in https://github.com/knative/pkg/tree/master/configmap/hash-gen which generates a hash from the content of ConfigMaps to prevent people from accidentally modifying `_example` values. Whenever a change is performed to `_example` data, the hash has to be re-generated.

Because those modifications happened in the past, I would refrain from touching the historical manifests saved in the operator repo (therefore marking as "do not merge"), and instead fix the typos directly inside the serving and eventing repos so we don't carry them here in the future.

cc @houshengbo

/hold


jsoref:
Thanks @antoineco

PR #309, opened 2020-10-19
Consolidate the used YAML dependencies to just one

markusthoemmes:
## Proposed Changes

I opted to go with `sigs.k8s.io/yaml` because... it's from K8s and we're mangling K8s stuff so :man_shrugging: 

**Release Note**

```release-note
NONE
```

/assign @houshengbo


matzew (approved):
/lgtm
/approve

PR #312, opened 2020-10-21
Add ingress configuration support

### LINKED ISSUE #64: Serving Operator should provide an option that all ingress gateways are configured
If a user creates a custom ingress gateway, istio may select a gateway at random (https://github.com/knative/serving/issues/5163#issuecomment-521815915). The operator should provide a way to ensure that the ingress gateway always works.

This relies a bit on https://github.com/knative/serving/issues/5172.

I believe this could work by montoring gateways and adjusting the config-istio to account for the new gateway (https://github.com/knative/serving/blob/b7dfe3fdb80ffac7f7a05c20787400ce51aad2b1/config/config-istio.yaml#L45)

This issue is a clone of an existing issue in knative/serving-operator: https://github.com/knative/serving-operator/issues/156

houshengbo:
Co-authored-by: Markus Thömmes \<mthoemme@redhat.com\>

Fixes #64 

## Proposed Changes

* Added Kourier, Contour and Istio as the available ingresses.
* Read the ingress manifests under kodata/ingress/\<version\>.
* Serving component aligns the ingress at the same minor version.

**Release Note**

```release-note

```


nak3 (commented):
There are still many of Istio specific code like:
https://github.com/knative/operator/blob/18aca414e5109ee3f66b4431430616c65ef799ab/pkg/reconciler/knativeserving/common/ha.go#L38
https://github.com/knative/operator/blob/18aca414e5109ee3f66b4431430616c65ef799ab/pkg/reconciler/knativeserving/common/gateway.go#L26
It would be a problem if we use non-istio Ingress, wouldn't it?

markusthoemmes:
Has the feature doc been updated to reflect the necessary changes and have those been discussed and accepted?

pkg/reconciler/common/releases.go
```diff
@@ -221,6 +221,33 @@ func createManifestsPath(instance v1alpha1.KComponent) []string {
 	return nil
 }
 
+// componentIngressPath composes the ingress path for serving based on version
+func componentIngressPath(version string, instance v1alpha1.KComponent) string {
+	koDataDir := os.Getenv(KoEnvKey)
+	switch instance.(type) {
+	case *v1alpha1.KnativeServing:
```
jcrossley3, line 228: It's surprising to see any mention of "ingress" in the common package since it's strictly related to `KnativeServing`. I wouldn't expect any function operating on a `KComponent` to know anything about an ingress.

houshengbo: I moved the functions into ingress package. They are called by the knativeserving reconciler.

pkg/apis/operator/v1alpha1/knativeserving_types.go
```diff
@@ -111,3 +114,25 @@ type HighAvailability struct {
 	// will be scaled to.
 	Replicas int32 `json:"replicas"`
 }
+
+// IngressConfigurations specifies options for the ingresses.
+type IngressConfigurations struct {
+	Istio   IstioIngressConfiguration   `json:"istio"`
+	Kourier KourierIngressConfiguration `json:"kourier"`
+	Contour ContourIngressConfiguration `json:"Contour"`
```
jcrossley3, line 122:
```suggestion
	Contour ContourIngressConfiguration `json:"contour"`
```

pkg/reconciler/knativeserving/ingress/ingress.go
```diff
@@ -0,0 +1,126 @@
...
+	"knative.dev/operator/pkg/reconciler/common"
+)
+
+const providerLabel = "networking.knative.dev/ingress-provider"
+
+func ingressFilter(name string) mf.Predicate {
+	return func(u *unstructured.Unstructured) bool {
+		provider, hasLabel := u.GetLabels()[providerLabel]
+		if !hasLabel {
+			return true
```
nak3, line 37: I might be wrong, but is this `return true` correct? If the resources have no `providerLabel`, operator manages it? :thinking:

nak3 (commented):
`config/300-serving.yaml` also needs to be updated?

pkg/reconciler/knativeserving/ingress/ingress.go
houshengbo: If the providerLabel is NOT found, the resource is then from the knative serving, so we need to keep it by returning true. :-)

evankanderson:
Having just checked in #360 , I wonder whether we'd prefer to simply ship different config directories for serving + each ingress plugin.

My main concern having put that together was making sure that if you tried to pin to a particular release, you'd get the same manifests even if you upgraded the operator, by using the logic here:

https://github.com/knative/operator/blob/fd53bd554e19788c081861f1c9206e059a7c053c/pkg/packages/release.go#L170

Sadly, I forgot to write a comment; but the logic for "additional packages" is to use the most-recent release of the additional package which is _older_ than the selected patch version, if one exists, otherwise to use the ".0" patch-release. This does mean that the main package might need to create a new patch release if it wants to expose a new patch release in an additional component, but provides a repeatable way (assuming forward flow of time) to align the manifests.

evankanderson:
@markusthoemmes -- I'd love your thoughts on using "latest minor" for additional packages or "snapshotted files" like I suggested in #360. A sample config if we wanted to use the fetcher to maintain alternate ingress configs would look like this gist:

https://gist.github.com/evankanderson/45fc3d45f487c2de77b0da5b373d5f29

evankanderson:
(also @houshengbo , but I was chatting with him on slack)

houshengbo:
@evankanderson @markusthoemmes 
PLZ check the design doc: https://docs.google.com/document/d/1AQBGCIoUAB8Whg02MVWdujMj8O-OWFvPbccoctMbgJA/edit#
I pasted a snapshot, regarding the structure of the ingress directory in the kodata. That is also what I am implementing in this PR.

About how to align the versions of knative serving and ingress, the answer from me is that we align by the MINOR version.
When user specifies `ma.mi.pa` or `ma.mi` in the spec.version field, operator just picks up the value of `ma.mi`, and locates the `ma.mi` folder under ingress to load all the ingress manifests.

The directory ingress/`ma.mi` saves the LATEST available ingress version available under the MINOR version. We can update it from time to time, when there is new patch release for a certain ingress.

pkg/apis/operator/v1alpha1/knativeserving_types.go
```diff
@@ -111,3 +114,43 @@ type HighAvailability struct {
...
+
+// KongIngressConfiguration specifies whether to enable the kong ingresses.
+type KongIngressConfiguration struct {
+	Enabled bool `json:"enabled"`
+}
+
+// GlooIngressConfiguration specifies whether to enable the gloo ingresses.
+type GlooIngressConfiguration struct {
+	Enabled bool `json:"enabled"`
+}
```
markusthoemmes, line 156: Let's drop this as long as we don't ship the manifests.

pkg/reconciler/knativeserving/ingress/kourier.go
```diff
@@ -0,0 +1,67 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes/scheme"
+
+	mf "github.com/manifestival/manifestival"
+	"knative.dev/operator/pkg/apis/operator/v1alpha1"
+)
+
+const KourierGatewayNSEnvVarKey = "KOURIER_GATEWAY_NAMESPACE"
```
markusthoemmes, line 30: No need to make this globally available I think. Private (lowercase) should suffice.

pkg/reconciler/knativeserving/ingress/testdata/kodata/ingress/0.18/kourier.yaml
```diff
@@ -0,0 +1,98 @@
+apiVersion: apps/v1
```
markusthoemmes, line 1: Should these be symlinks?

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -113,27 +116,33 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *v1alpha1.KnativeServ
 	return stages.Execute(ctx, &manifest, ks)
 }
 
+// enableIngresses removes the disabled ingresses from the manifests
+func (r *Reconciler) enableIngresses(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent) error {
```
markusthoemmes, line 120: Can we come up with a better name? Like `filterDisabledIngresses`?

pkg/reconciler/knativeserving/ingress/ingress_test.go
```diff
@@ -0,0 +1,414 @@
...
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	mf "github.com/manifestival/manifestival"
+	servingv1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
+	"knative.dev/operator/pkg/reconciler/common"
+	util "knative.dev/operator/pkg/reconciler/common/testing"
+)
+
+const NumberIngressResource = 5
```
markusthoemmes, line 36: No need for public visibility

pkg/reconciler/knativeserving/ingress/ingress_test.go
```diff
@@ -0,0 +1,414 @@
...
+
+func TestAppendInstalledIngresses(t *testing.T) {
+	os.Setenv(common.KoEnvKey, "testdata/kodata")
+	defer os.Unsetenv(common.KoEnvKey)
+
+	tests := []struct {
+		name                 string
+		instance             servingv1alpha1.KnativeServing
+		expected             bool
+		expectedIngressesNum int
```
markusthoemmes, line 82: I don't quite understand what this checks. Can we make the test explicitly check that only istio is loaded by default, Kourier if specified and Istio AND Kourier if both are specified?

markusthoemmes (commented):
Looks good overall, but I'd tune down the number of supported Ingresses to just the number we ship, potentially even to just 2 to make the diff quite a bit smaller and more focused.

pkg/apis/operator/v1alpha1/knativeserving_types.go
```diff
@@ -64,6 +64,9 @@ type KnativeServingSpec struct {
 	// Allows specification of HA control plane
 	// +optional
 	HighAvailability *HighAvailability `json:"high-availability,omitempty"`
+
+	// Ingress allows configuration of different ingress adapters to be shipped.
+	Ingress *IngressConfigurations `json:"ingress,omitempty"`
```
maximilien, line 69: Would `IngressConfigs` be a better name?

pkg/reconciler/knativeserving/ingress/gloo.go
```diff
@@ -0,0 +1,30 @@
...
+	"context"
+
+	mf "github.com/manifestival/manifestival"
+	"knative.dev/operator/pkg/apis/operator/v1alpha1"
+)
+
+var glooFilter = ingressFilter("gloo")
+
+func glooTransformers(ctx context.Context, instance *v1alpha1.KnativeServing) []mf.Transformer {
+	return nil
```
maximilien, line 29: Interesting you return `nil` here and not an empty `[]mf.Transformer{}`...

pkg/reconciler/knativeserving/ingress/gloo_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+
+	servingv1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
+	util "knative.dev/operator/pkg/reconciler/common/testing"
+)
+
+func TestGlooTransformers(t *testing.T) {
+	instance := &servingv1alpha1.KnativeServing{}
+	transformer := glooTransformers(context.TODO(), instance)
+	util.AssertEqual(t, len(transformer), 0)
```
maximilien, line 30: This should fail if nil is returned from code... no?

pkg/reconciler/knativeserving/ingress/kong.go
```diff
@@ -0,0 +1,30 @@
...
+	"context"
+
+	mf "github.com/manifestival/manifestival"
+	"knative.dev/operator/pkg/apis/operator/v1alpha1"
+)
+
+var kongFilter = ingressFilter("kong")
+
+func kongTransformers(ctx context.Context, instance *v1alpha1.KnativeServing) []mf.Transformer {
+	return nil
```
maximilien, line 29: same as for the gloo transformer

pkg/reconciler/knativeserving/ingress/testdata/kodata/ingress/0.18/kourier.yaml
maximilien: Would love that too 👍

maximilien (approved):
Left a few comments... still learning this codebase.

pkg/reconciler/knativeserving/ingress/gloo.go
houshengbo: If it is nil, then the length of the array is 0.

pkg/reconciler/knativeserving/ingress/gloo_test.go
houshengbo: This test case is removed.

pkg/reconciler/knativeserving/ingress/gloo.go
markusthoemmes: Idiomatic Golang allows for transparent use of `nil` slices. Their len is 0 and you can `range` over them no problem. Allows for saving the allocations that returning an empty slice forces.

pkg/reconciler/knativeserving/ingress/ingress_test.go
houshengbo: This test cases check whether it can get exactly the same number of resources available in the ingress directory. I change the name into `expectedResourcesNum`.
To verify if several specified ingresses are loaded, I will add them into the test case `TestTransformers`.

nak3:
I tested with Kourier but it still does not work if I deployed operator in non-`knative-serving` namespace.
The reason is that Kourier gateway find the controller by the `address` in kourier-bootstrap CM:

https://github.com/knative/operator/pull/312/files#diff-0aac84f00c17a91f67fcac679f691c8ac69fadf9379b44070a6b6fa4843cfaeaR352
```
address: "kourier-control.knative-serving"    # The `knative-serving` is a namespace of kourier-control.
````

So the address needs to be updated somehow.

**Comment by nak3** [comment]
(With said, I think this needs to be fixed in Kourier. It should be configurable by env value or something easier.)

File: pkg/reconciler/knativeserving/ingress/kourier.go [hunk 14]
```diff
@@ -0,0 +1,67 @@
...
+	return []mf.Transformer{
+		replaceKourierGWNamespace(instance.GetNamespace()),
+	}
+}
+
+// replaceKourierGWNamespace replace the environment variable KOURIER_GATEWAY_NAMESPACE with the namespace of the Knative Serving CR
+func replaceKourierGWNamespace(ns string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() == "Deployment" {
+			_, hasLabel := u.GetLabels()[providerLabel]
```
markusthoemmes, line 44: This filter seems odd. Shouldn't this match a very specific deployment instead?

markusthoemmes (commented):
Agree with @nak3. We should invest in writing a transform that mucks around with the config potentially. We can ship that after this initially lands.

This is why I wanted to just do **one** ingress at first. To not block things too much on specialties of the other ingresses and to just land the concept.

houshengbo:
@markusthoemmes @nak3 It is possible to write a transformer working on the ConfigMap for Kourier, but would the CM like `kourier-bootstrap` have a fixed structure for data section? It looks like we need dig into multiple layers down to locate `envoy-bootstrap.yaml.static_resources.clusters[1].load_assignment.......address`.
 Operator supports the configuration directly on ConfigMap, with `spec.config` in the CR. If there is a transformer for the ConfigMap, it may lead to some conflict, when we leverage `spec.config` to overwrite the ConfigMap.

houshengbo:
I opened another issue to track Kourier ns: https://github.com/knative/operator/issues/454

nak3:
@houshengbo @markusthoemmes  Thank you! 

One more question, do you think that we don't need to add [contour.yaml](https://github.com/knative-sandbox/net-contour/tree/master/config/contour) for Contour?
At first, I thought that users should install Contour from net-contour or projectcontour as they like (it is same manner with Istio), but Contour's [this job](https://github.com/knative-sandbox/net-contour/blob/release-0.20/config/contour/external.yaml#L2054) is net-contour specific. So users have to deploy the `contour.yaml` from net-contour repo even after enabled contour.

houshengbo:
@nak3 My initial thought is that we stay with net-***.yaml ingress plugin only in operation. Users need to install the ingress by themselves, like istio.
However, Kourier is an exception. They do not need to install Kourier separately, since kourier.yaml has everything in operator, and it is the only artifact we ship. 
It seems now the job you mentioned is a bit awkward. Why is it in the contour.yaml, not net-contour.yaml?
@markusthoemmes Yours?

nak3:
Sorry the job is not net-contour specific. It was misunderstanding.

But it is still so difficult to deploy without using "net-contour's" Contour. net-contour has [`release.yaml`](https://github.com/knative-sandbox/net-contour/releases) which includes all of them so I wonder that it is better to use them.
With said, I understood what you wanted to do. Hmm... :thinking:

houshengbo:
@nak3 If we use the release.yaml consolidating everything of contour, users do not need to care about installing contour? Is it correct?
What if they have already installed contour? Would there be conflict?

nak3:
Alright, LGTM then. I am leaving other reviewers.

markusthoemmes (approved):
/lgtm
/approve

PR #319, opened 2020-10-23
Add validation to the field spec.version

houshengbo:
Target version should be in ma.mi or ma.mi.pa.
Current status version should be in ma.mi.pa.

Fixes #

## Proposed Changes

*
*
*

**Release Note**

```release-note

```


pkg/reconciler/common/releases.go
```diff
@@ -74,42 +74,53 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
...
 	if current == "" {
-		return true
+		return nil
 	}
-	current = sanitizeSemver(current)
-	target := sanitizeSemver(TargetVersion(instance))
 
+	current = sanitizeSemver(current)
+	errMess = fmt.Errorf("not supported to upgrade or downgrade across multiple MINOR versions. The "+
+		"installed KnativeServing version is %v.", current)
```
antoineco, line 103: I find it a bit confusing to see an error initialized even before the comparison is even performed. Maybe move this inside the `if` block below?

pkg/reconciler/common/releases.go
```diff
@@ -74,42 +74,53 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
...
+		return fmt.Errorf("target version %v should at least include the major and minor numbers.", target)
+	}
+
+	targetMinor, err := strconv.Atoi(strings.Split(target, ".")[1])
+	if err != nil {
+		return fmt.Errorf("minor number of the target version %v should be an integer.", target)
+	}
+
 	current := instance.GetStatus().GetVersion()
 	// If there is no manifest installed, return true, because the target manifest is able to install.
```
antoineco, line 96: The returned value is not a bool anymore.
```suggestion
	// If there is no manifest installed, return because the target manifest is able to install.
```

pkg/reconciler/common/releases.go
```diff
@@ -74,42 +74,53 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
 	return fetch(installedManifestPath(current, instance))
 }
 
-// IsUpDowngradeEligible returns the bool indicate whether the installed manifest is able to upgrade or downgrade to
-// the target manifest.
-func IsUpDowngradeEligible(instance v1alpha1.KComponent) bool {
+// IsVersionValidMigrationEligible returns the bool indicate whether the target version is valid and the installed
+// manifest is able to upgrade or downgrade to the target manifest.
+func IsVersionValidMigrationEligible(instance v1alpha1.KComponent) error {
+	var errMess error
```
antoineco, line 80: Since this is an actual error interface, and not a message, why not sticking to a traditional `err` variable name?

pkg/reconciler/common/releases.go
```diff
@@ -74,42 +74,53 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
...
+	}
+
+	if len(strings.Split(target, ".")) < 2 {
+		return fmt.Errorf("target version %v should at least include the major and minor numbers.", target)
+	}
+
+	targetMinor, err := strconv.Atoi(strings.Split(target, ".")[1])
+	if err != nil {
+		return fmt.Errorf("minor number of the target version %v should be an integer.", target)
+	}
```
antoineco, line 93: If not sure if I understand that part. Why only checking the minor, and not the major and patch? Wouldn't the `semver` lib be able to validate the whole Semver string here?

pkg/reconciler/common/releases.go
```diff
@@ -74,42 +74,53 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
...
-		return false
-	}
-
-	targetMinor, err := strconv.Atoi(strings.Split(target, ".")[1])
-	if err != nil {
-		return false
+		return fmt.Errorf("minor number of the current version %v should be an integer.", current)
 	}
 
 	// If the diff between minor versions are less than 2, return true.
```
antoineco, line 118: Not a bool anymore.
```suggestion
	// If the diff between minor versions are less than 2, return.
```

pkg/reconciler/common/releases.go
houshengbo: there are multiple errors in this function.

pkg/reconciler/common/releases.go
houshengbo: We check the major number first, then the minor.
these lines of code should be located downstairs.
i moved then upwards because i need to validate the target version.

pkg/reconciler/common/releases.go
antoineco: Sure, but they could all use the same variable from what I see. My point is that having both `errMss` and `err` within the same function doesn't really help understanding which one is used for what purpose (as opposed to something called `validationErr`, for example).

pkg/reconciler/common/releases.go
antoineco: You're checking the minor here, not the major. The logic is

1. check minor
1. do something unrelated
1. check major

But why? I still don't get it.

antoineco (approved):
/lgtm

PR #326, opened 2020-10-26
Add support of the major.minor format for spec.version

### LINKED ISSUE #316: Support the MAJOR.MINOR format for the value of `spec.version`
We support the format of MAJOR.MINOR.PATCH for the field `spec.version`.
As we need to align serving and eventing components with other components at the MINOR number, it is convenient for the operator to support the MAJOR.MINOR format for the field `spec.version`.

houshengbo:
Fixes #316

### DIFF
--- a/pkg/reconciler/common/releases.go
+++ b/pkg/reconciler/common/releases.go
@@ -46,15 +46,18 @@ var cache = map[string]mf.Manifest{}
 // per the spec in the component. If spec.version is empty, the latest
 // version known to the operator is returned.
 func TargetVersion(instance v1alpha1.KComponent) string {
-	target := instance.GetSpec().GetVersion()
-	if target == "" {
-		// If spec.version is not set and spec.manifests are set, we leave the target version empty.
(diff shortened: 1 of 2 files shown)


pkg/reconciler/common/releases.go
```diff
@@ -291,10 +294,23 @@ func allReleases(instance v1alpha1.KComponent) ([]string, error) {
...
+func getLatestRelease(instance v1alpha1.KComponent, version string) string {
+	// The versions are in a descending order, so the first one will be the latest version.
 	vers, err := allReleases(instance)
 	if err != nil {
 		panic(err)
 	}
-	// The versions are in a descending order, so the first one will be the latest version.
-	return vers[0]
+	for _, val := range vers {
+		if strings.HasPrefix(val, version) {
```
Cynocracy, line 309: I think here we want version + '.' or we'll match incorrectly in some cases

for ex searching for version 1 might match 11 if there isn't an actual 1.0

houshengbo: oh, that's correct.

PR #349, opened 2020-11-11
Add namespace injection into the manifests in upgrade tests

houshengbo:
## Proposed Changes

* In test cases, if ns is the only difference between two resources, we consider them identical. The namespace is injected into the manifests, so that ns difference won't be taken as resource difference.


Cynocracy (commented):
/lgtm

PR #360, opened 2020-11-20
Add a tool to automatically fetch releases and stage them in `kodata`

evankanderson:
This needs more (any!) tests, but it produces a small diff and I wanted to see what people thought of the design while I produced tests.

## Proposed Changes

* Automate the creation and management of `kodata` files for the release.
* **NOTE:** this does not hook the generation into `update-codegen` at the moment, but that would be a logical next step.


evankanderson:
Note that this looks enormous, but it's just because of https://github.com/knative/operator/pull/360/commits/7fb9fd03530adcc8b974db6aeb8e64f087883cd7, which pulls in 4 or 5 releases.

evankanderson:
Of greater interest, there are different SHA1 checksums for the net-istio packages in 0.17.3 and 0.17.4 serving.

https://github.com/knative/operator/pull/360/commits/7fb9fd03530adcc8b974db6aeb8e64f087883cd7#diff-9bb746ae275a424dbbf77648bad82f6f129c5a8ceeade8f618345fc876861704

houshengbo:
There used to be bugs in 0.17.x, so I have to manually align the patch number of net-istio with the knative-serving, but I do not think we have to do that for current version.

houshengbo:
In general, I love this tool. We probably need to evolve it, once we implement features like multi-ingress support, as the kodata structure may change.
Do we have a automated job that runs the tool, when a new release is detected?

evankanderson:
If we make this part of update-codegen, it should run nightly the same way that update-deps does (and send PRs when we need to check in new files).

My not-so-secret plan is also to provide a daemon mode which could be used as a sidecar to the current controller to populate the `kodata` directory for the main controller.

Now with more tests.

evankanderson:
It looks like I need to boost coverage a little more on release.go, I'll see what I can do.

evankanderson:
@houshengbo 

Updated; once this is in, I'll work on autogen so that you don't need to keep doing this by hand.

PR #379, opened 2020-12-07
Add the field spec.additionalManifests into the CRD's

### LINKED ISSUE #330: Allow users to define additional manifests without wiping the Serving/Eventing core
**Problem**

After defining a custom `manifest` in my KnativeServing object, the entire core Serving landcape was wiped.

I was expecting the operator to deploy my manifest _additionally_ to the core. I think most people do want to have the Knative core when using the operator, so this behaviour was a bit surprising to me, and not that easy to recover from in a live cluster.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** Cluster admins

**Exit Criteria**

Custom manifests can be deployed _on top_ of the core, without having to reference every core manifests from GitHub.

If we don't want to break the current functionality, this will require a new option. I suggest `additionalManifests`.

**Time Estimate (optional):** ?

cc. @houshengbo as discussed in Slack

houshengbo:
Fixes #330

## Proposed Changes

* This PR changes the structure of the CRD's by adding the spec.additionalManifests.

### DIFF
--- a/config/300-eventing.yaml
+++ b/config/300-eventing.yaml
@@ -71,6 +71,15 @@ spec:
                   URL:
                     description: The link of the manifest URL
                     type: string
+            additionalManifests:
+              description: A list of the additional eventing manifests, which will be installed by the operator
+              type: array
+              items:
+                type: object
+                properties:
+                  URL:
+                    description: The link of the additional manifest URL
+                    type: string
             config:
               additionalProperties:
                 additionalProperties:
--- a/pkg/reconciler/common/stages.go
+++ b/pkg/reconciler/common/stages.go
@@ -83,7 +83,9 @@ type ManifestFetcher func(ctx context.Context, instance v1alpha1.KComponent) (*m
 // manifest is captured in a closure before any stage might mutate the
 // instance status, e.g. Install.
 func DeleteObsoleteResources(ctx context.Context, instance v1alpha1.KComponent, fetch ManifestFetcher) Stage {
-	if TargetVersion(instance) == instance.GetStatus().GetVersion() {
+	version := TargetVersion(instance)
+	if version == instance.GetStatus().GetVersion() && targetManifestPath(version,
+		instance) == installedManifestPath(version, instance) {
 		return NoOp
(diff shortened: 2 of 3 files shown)


pkg/reconciler/common/stages.go
```diff
@@ -83,7 +83,9 @@ type ManifestFetcher func(ctx context.Context, instance v1alpha1.KComponent) (*m
 // manifest is captured in a closure before any stage might mutate the
 // instance status, e.g. Install.
 func DeleteObsoleteResources(ctx context.Context, instance v1alpha1.KComponent, fetch ManifestFetcher) Stage {
-	if TargetVersion(instance) == instance.GetStatus().GetVersion() {
+	version := TargetVersion(instance)
+	if version == instance.GetStatus().GetVersion() && targetManifestPath(version,
+		instance) == installedManifestPath(version, instance) {
```
antoineco, line 88: Maybe use a more idiomatic style of line wrapping here?
```suggestion
	if version == instance.GetStatus().GetVersion() &&
		targetManifestPath(version, instance) == installedManifestPath(version, instance) {

```

config/300-eventing.yaml
```diff
@@ -71,6 +71,15 @@ spec:
...
                     type: string
+            additionalManifests:
+              description: A list of the additional eventing manifests, which will be installed by the operator
+              type: array
+              items:
+                type: object
+                properties:
+                  URL:
+                    description: The link of the additional manifest URL
+                    type: string
```
antoineco, line 82: Just a suggestion but we could even declare an expected `format`, since OpenAPI supports it? I don't think it is taken into account during validation, but it's a nice hint I think. Up to you, feel free to reject the suggestion.
```suggestion
                    type: string
                    format: uri
```

houshengbo: We actually validate the resources, when using the manifestival library to retrieve them by the `spec.manifests`.
No need to worry about the format of this field.

antoineco (approved):
/lgtm

PR #386, opened 2020-12-10
Point all dependencies to 0.19 branches properly

markusthoemmes:
The 0.19 release is currently pointing at a much newer `pkg` than is on the release-0.19 branch. That breaks downstreams that depend on exact 0.19 coordinates and is plainly wrong anyway.

Fixed that by running `./hack/update-deps.sh --upgrade --release 0.19`


matzew (approved):
/lgtm
/approve

PR #394, opened 2020-12-14
Preserve the values of the environment variables for the pingsource-mt-adapter deploy resource

### LINKED ISSUE #393: Env var changes of "pingsource-mt-adapter" are not respected
https://github.com/knative/operator/commit/eaf60c26acaa42995ff456116a274e382c5d886c introduced some ignoring logic around replicas and env vars for the `pingsource-mt-adapter`.

Any changes to env vars (be it additions or changes to existing variables) will be ignored for existing deployments, which might come around to bite us badly.

houshengbo:
Fixes #393

### DIFF
--- a/pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
+++ b/pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
@@ -24,6 +24,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/kubernetes/scheme"
+	"knative.dev/pkg/system"
+)
+
+var (
+	envVarNames = []string{system.NamespaceEnvKey, "K_METRICS_CONFIG", "K_LOGGING_CONFIG", "K_LEADER_ELECTION_CONFIG",
+		"K_NO_SHUTDOWN_AFTER", "K_SINK_TIMEOUT"}
 )
 
 type unstructuredGetter interface {
@@ -63,9 +69,15 @@ func ReplicasEnvVarsTransform(client unstructuredGetter) mf.Transformer {
 					applyContainer := &apply.Spec.Template.Spec.Containers[containerIndex]
 					mergedEnv := currentContainer.Env
 					for _, env := range applyContainer.Env {
-						if !nameExistsEnvVars(env.Name, mergedEnv) {
-							// Add the new env var into the existing env vars
+						found, envIndex := nameExistsEnvVars(env.Name, mergedEnv)
+						if !found {
+							// Add the new env var into the existing env vars, if it is not available in the existing
+							// cluster.
 							mergedEnv = append(mergedEnv, env)
+						} else if !nameExistsEnvVarNames(env.Name, envVarNames) {
+							// Set the env var value, if it is available in the existing
+							// cluster, but it is not in the preserved list of the env vars.
+							mergedEnv[envIndex] = env
 						}
 					}
 					applyContainer.Env = mergedEnv
@@ -83,9 +95,18 @@ func ReplicasEnvVarsTransform(client unstructuredGetter) mf.Transformer {
 	}
 }
 
-func nameExistsEnvVars(name string, envvars []corev1.EnvVar) bool {
-	for _, env := range envvars {
+func nameExistsEnvVars(name string, envvars []corev1.EnvVar) (bool, int) {
+	for index, env := range envvars {
 		if env.Name == name {
+			return true, index
+		}
+	}
+	return false, -1
+}
+
+func nameExistsEnvVarNames(name string, envVarNames []string) bool {
+	for _, envVarName := range envVarNames {
+		if envVarName == name {
 			return true
 		}
 	}
--- a/pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
+++ b/pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
@@ -19,6 +19,7 @@ package common
 import (
 	mf "github.com/manifestival/manifestival"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -45,6 +46,7 @@ func ReplicasEnvVarsTransform(client unstructuredGetter) mf.Transformer {
 			if err := scheme.Scheme.Convert(u, apply, nil); err != nil {
 				return err
 			}
+
 			current := &appsv1.Deployment{}
 			if err := scheme.Scheme.Convert(currentU, current, nil); err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


markusthoemmes:
What about changing existing env vars that are not overridden downstream? Should we instead opt to hardcode which env vars to leave alone?

lionelvillard:
I agree with @markusthoemmes. Let's hardcode the env vars that are managed by the controller (see https://github.com/knative/eventing/blob/master/pkg/reconciler/pingsource/resources/receive_adapter.go#L37), until we find a better solution.

matzew:
yeah, I was thinking same, have an hardcoded list of existing ones.... for now...

and let's have an issue for something better?

pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
```diff
@@ -19,10 +19,17 @@ package common
...
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/kubernetes/scheme"
+	"knative.dev/pkg/system"
+)
+
+var (
+	envVarNames = []string{system.NamespaceEnvKey, "K_METRICS_CONFIG", "K_LOGGING_CONFIG", "K_LEADER_ELECTION_CONFIG",
```
markusthoemmes, line 31: make this a `sets.String` to simplify the functions below (you won't need a find function anymore, as a set has a `Has` method)

pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
```diff
@@ -53,13 +61,26 @@ func ReplicasEnvVarsTransform(client unstructuredGetter) mf.Transformer {
...
+						found, envIndex := nameExistsEnvVars(env.Name, mergedEnv)
+						if !found {
+							// Add the new env var into the existing env vars, if it is not available in the existing
+							// cluster.
+							mergedEnv = append(mergedEnv, env)
+						} else if !nameExistsEnvVarNames(env.Name, envVarNames) {
+							// Set the env var value, if it is available in the existing
+							// cluster, but it is not in the preserved list of the env vars.
+							mergedEnv[envIndex] = env
+						}
```
markusthoemmes, line 81: Should we be handling deletes of env vars that are no longer in the manifest too?

houshengbo: Good question, I need to ask @lionelvillard or @matzew

markusthoemmes: That was a rhetoric question. I think we should handle that :)

pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
```diff
@@ -74,13 +75,23 @@ func ReplicasEnvVarsTransform(client unstructuredGetter) mf.Transformer {
...
 							// Set the env var value, if it is available in the existing
 							// cluster, but it is not in the preserved list of the env vars.
 							mergedEnv[envIndex] = env
 						}
 					}
-					applyContainer.Env = mergedEnv
+
+					// Remove the env var, which is not in the preserved list and not in the applied manifests
+					cleanedMergedEnv := mergedEnv
+					for _, env := range mergedEnv {
```
markusthoemmes, line 87: hmm, I don't think this is necessary at is.

What do you think about only iterating over the existing deployment once, copying all `envVarNames` vars into a `[]corev1.EnvVar` and then appending all env vars not  in `envVarNames` from the `applyContainer`, like

```golang
var vars []corev1.EnvVar

for _, env := range currentContainer.Env {
  if envVarNames.Has(env.Name) {
    vars = append(vars, env)
  }
}

for _, env := range applyContainer.Env {
  if envVarNames.Has(env.Name) {
    vars = append(vars, env)
  }
}

applyContainer.Env = vars
```

That should do everything the current code does, I believe.

houshengbo: Something is missing: if the env belongs to the preserved list, but the current does not have this env, we SHOULD apply this env.

pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
```diff
@@ -53,13 +62,31 @@ func ReplicasEnvVarsTransform(client unstructuredGetter) mf.Transformer {
...
+							// Keep all the env vars in the preserved list
+							mergedEnv = append(mergedEnv, env)
+						}
 					}
+					for _, env := range applyContainer.Env {
+						if !envVarNames.Has(env.Name) {
+							// If the env var in the applied container is not in the preserved list,
+							// we should apply it.
+							mergedEnv = append(mergedEnv, env)
+						} else if !nameExistsEnvVars(env.Name, currentContainer.Env) {
```
markusthoemmes, line 83: Why is this path necessary? We start fresh and have a clear preserve list, which we exclude already. We should append all other env vars here I believe.

houshengbo: if the env belongs to the preserved list, the current does not have this env, but the apply container has it, we SHOULD apply this env. No?

markusthoemmes: GREAT point! I guess the more performant way of doing that is to build up a new `sets.String` in the first loop, adding every variable that was **actually** there, and then check against that new list in the second loop. Spares you the quad complexity. Somthing like

```golang
actualKeys := sets.NewString()
for _, env := range currentContainer.Env {
    if envVarNames.Has(env.Name) {
        // Keep all the env vars in the preserved list
        mergedEnv = append(mergedEnv, env)
        actualKeys.Insert(env.Name)
    }
}

for _, env := range applyContainer.Env {
    if !actualKeys.Has(env.Name) {
        // Apply all keys that are neither preserved, nor in the actual container.
        mergedEnv = append(mergedEnv, env)
    }
}
```

I might be halluzinating though, since I'm super hungry :joy:. Feel free to ignore this comment too :), the performance implications aren't going to matter anyway.

houshengbo: Feel free to call it a day. Thx for the comments.

pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
matzew: > Should we be handling deletes of env vars that are no longer in the manifest too?

yes

houshengbo: We have added the code to handle the deletion.

lionelvillard:
LGTM

@markusthoemmes for final stamp

pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
```diff
@@ -73,3 +99,12 @@ func ReplicasEnvVarsTransform(client unstructuredGetter) mf.Transformer {
 		return nil
 	}
 }
+
+func nameExistsContainers(name string, containers []corev1.Container) (bool, int) {
```
markusthoemmes, line 103: I'd make this a `findContainer(name string, containers []corev1.Container) *corev1.Container` to give it a more speaking name and avoid the index shenanigans.

pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
```diff
@@ -53,13 +62,30 @@ func ReplicasEnvVarsTransform(client unstructuredGetter) mf.Transformer {
...
 
+			// Preserve the env vars in the existing cluster
 			for index := range current.Spec.Template.Spec.Containers {
 				currentContainer := current.Spec.Template.Spec.Containers[index]
-				for j := range apply.Spec.Template.Spec.Containers {
-					applyContainer := &apply.Spec.Template.Spec.Containers[j]
-					if currentContainer.Name == applyContainer.Name {
-						applyContainer.Env = currentContainer.Env
+				found, containerIndex := nameExistsContainers(currentContainer.Name, apply.Spec.Template.Spec.Containers)
+				if found {
```
markusthoemmes, line 69: You can spare yourself a level of indentation if you do 

```golang
if !found {
  continue
}

applyContainer := ...
```

markusthoemmes (commented):
Two nits to increase readability, then it's LGTM I think.

pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
houshengbo: I changed the name and the way to implement.

pkg/reconciler/knativeeventing/common/replicasenvvarstransform.go
houshengbo: As I re-implement `findContainer(name string, containers []corev1.Container) *corev1.Container`, we can determine the container is found or not, by checking if the returned container is nil or not.

markusthoemmes (approved):
/lgtm
/approve

PR #409, opened 2020-12-16
Add the missing serving and eventing manifests

houshengbo:


antoineco:
Seems like most of it was already merged somewhere else.
/lgtm

PR #430, opened 2021-01-15
Refactor the function getLatestRelease and add the correct error message

houshengbo:
## Proposed Changes

* The func getLatestRelease returns spec.version, if spec.version is not available.
* Add error message into the CR, if the manifest is not available.

### DIFF
--- a/pkg/reconciler/common/releases.go
+++ b/pkg/reconciler/common/releases.go
@@ -82,6 +82,12 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
 func IsVersionValidMigrationEligible(instance v1alpha1.KComponent) error {
 	var err error
 	target := sanitizeSemver(TargetVersion(instance))
+	if target == "v" {
+		// If the function TargetVersion returns empty string, if means the specified target version is not available
+		// in this release to install.
+		return fmt.Errorf("target version %v is not available in this release.", instance.GetSpec().GetVersion())
+	}
+
 	if !semver.IsValid(target) {
 		return fmt.Errorf("target version %v is not in a valid semantic versioning format.", target)
 	}
--- a/pkg/reconciler/common/releases.go
+++ b/pkg/reconciler/common/releases.go
@@ -138,10 +138,15 @@ func getVersionKey(instance v1alpha1.KComponent) string {
 func getManifestWithVersionValidation(version string, instance v1alpha1.KComponent) (mf.Manifest, error) {
 	manifestsPath := targetManifestPath(version, instance)
 	manifests, err := fetch(manifestsPath)
-	if err != nil || (len(instance.GetSpec().GetManifests()) == 0 && len(instance.GetSpec().GetAdditionalManifests()) == 0) {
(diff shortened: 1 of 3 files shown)


pkg/reconciler/common/releases.go
```diff
@@ -82,6 +82,12 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
 func IsVersionValidMigrationEligible(instance v1alpha1.KComponent) error {
 	var err error
 	target := sanitizeSemver(TargetVersion(instance))
+	if target == "v" {
+		// If the function TargetVersion returns empty string, if means the specified target version is not available
+		// in this release to install.
+		return fmt.Errorf("target version %v is not available in this release.", instance.GetSpec().GetVersion())
+	}
```
antoineco, line 89: May I suggest this change to make things more readable? It feels to me like we should fail immediately after calling `TargetVersion()`, instead of passing something invalid to `sanitizeSemver()`, since we already know it's invalid at that point.

```suggestion
	trgtVer := TargetVersion(instance)
	if trgtVer == "" {
		// If the function TargetVersion returns empty string, if means the specified target version is not available
		// in this release to install.
		return fmt.Errorf("target version %v is not available in this release.", instance.GetSpec().GetVersion())
	}

	target := sanitizeSemver(trgtVer)
```

pkg/reconciler/common/releases.go
```diff
@@ -82,6 +82,12 @@ func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
 func IsVersionValidMigrationEligible(instance v1alpha1.KComponent) error {
 	var err error
 	target := sanitizeSemver(TargetVersion(instance))
+	if target == "v" {
+		// If the function TargetVersion returns empty string, if means the specified target version is not available
+		// in this release to install.
```
antoineco, line 87: This doesn't seem to be true, by looking at the code, it seems like `TargetVersion()` returns either:
* the latest available release
* `instance.GetSpec().GetVersion()`

If the version is not found, it returns `instance.GetSpec().GetVersion()`. Not `""`.

Could you please clarify the intention?

houshengbo: If `spec.version` is specified, but not available in the supported list of version numbers, TargetVersion returns "".
If `spec.version` is specified, and available in the supported list of version numbers, TargetVersion returns spec.version with the latest patch number.
If `spec.version` is NOT specified, TargetVersion returns the latest release number.

antoineco: Ok, I see the behavior now, somewhere deep inside `getLatestRelease()` (which name is misleading). IMO that code is very convoluted and should return an error if it can't return an actual version, but that's a topic for its own issue I think.

houshengbo: @antoineco I would like to change the logic for getLatestRelease.
There is an inconsistency between major.minor and major.minor.patch formats, if they are not found.
If getLatestRelease still returns the spec.version, no matter they are available or not. We can catch the error later, when parsing manifests.

antoineco (approved):
/lgtm

PR #443, opened 2021-01-26
Add `latest` as a special version supported by the operator

houshengbo:
## Proposed Changes

* This PR adds `latest` as a valid version Knative Operator supports with spec.version.


evankanderson (commented):
/lgtm

PR #462, opened 2021-02-11
Align all used YAML modules

markusthoemmes:
## Proposed Changes

As already done in other repos, this reduces the amount of YAML parser modules down to just one.

**Release Note**

```release-note
NONE
```

/assign @houshengbo @matzew


evankanderson:
When switching from `gopkg.in/yaml.v2` to `sigs.k8s.io/yaml`, you need to change the struct tags from `yaml:...` to `json:...`, as the Kubernetes library actually converts the YAML to JSON and then uses the JSON decoder library.

evankanderson:
(I have a PR for this shortly)

PR #468, opened 2021-02-12
Adds support for extension custom manifests

skonto:
## Proposed Changes

* Appends manifests defined in the controller extension into the main manifested loaded by the controller.
* Why? 
Currently downstream we have a scenario where a number of [custom monitoring resources](https://github.com/openshift-knative/serverless-operator/blob/400e76da629085e45f401b69dc3cbbf9900c376c/openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/0-rbac-proxy.yaml) are installed along with the 
core Serving ones. Although we can ship them together with the core ones, we would like to be able to create 
a number of manifests dynamically from a resource blueprint as there is a lot of repetition in the monitoring custom resources.
Another issue is that we need to filter core resources to apply specific transformations and this requires special transformation where we need to pick a resource based on some naming convention.
* Custom resources go through the validation process along with the rest of the manifests as usual assuming the appropriate label is defined. 
* Although there is support for custom manifests at the CRD level this does not really fit into our requirements for platform specific custom resources. Afaik, that functionality was meant for loading manifests from a Knative user perspective eg. ops persona. This patch serves the vendor persona.

/cc @markusthoemmes

### DIFF
--- a/pkg/reconciler/common/extensions.go
+++ b/pkg/reconciler/common/extensions.go
@@ -24,6 +24,7 @@ import (
 
 // Extension enables platform-specific features
 type Extension interface {
+	Manifests() []mf.Manifest
 	Transformers(v1alpha1.KComponent) []mf.Transformer
 	Reconcile(context.Context, v1alpha1.KComponent) error
 	Finalize(context.Context, v1alpha1.KComponent) error
@@ -39,6 +40,9 @@ func NoExtension(context.Context) Extension {
 
 type nilExtension struct{}
 
+func (nilExtension) Manifests() []mf.Manifest {
+	return nil
+}
 func (nilExtension) Transformers(v1alpha1.KComponent) []mf.Transformer {
 	return nil
 }
--- a/pkg/reconciler/common/extensions.go
+++ b/pkg/reconciler/common/extensions.go
@@ -24,7 +24,7 @@ import (
 
 // Extension enables platform-specific features
 type Extension interface {
-	Manifests() []mf.Manifest
+	Manifests(v1alpha1.KComponent) ([]mf.Manifest, error)
 	Transformers(v1alpha1.KComponent) []mf.Transformer
 	Reconcile(context.Context, v1alpha1.KComponent) error
 	Finalize(context.Context, v1alpha1.KComponent) error
@@ -40,8 +40,8 @@ func NoExtension(context.Context) Extension {
 
 type nilExtension struct{}
 
-func (nilExtension) Manifests() []mf.Manifest {
-	return nil
+func (nilExtension) Manifests(v1alpha1.KComponent) ([]mf.Manifest, error) {
+	return nil, nil
 }
 func (nilExtension) Transformers(v1alpha1.KComponent) []mf.Transformer {
 	return nil
--- a/pkg/reconciler/common/extensions_test.go
+++ b/pkg/reconciler/common/extensions_test.go
@@ -31,7 +31,7 @@ func (t TestExtension) Manifests() []mf.Manifest {
 	if err != nil {
 		return nil
 	}
-	return []mf.Manifest {manifest}
+	return []mf.Manifest{manifest}
 }
 
 func (t TestExtension) Transformers(v1alpha1.KComponent) []mf.Transformer {
--- a/pkg/reconciler/common/extensions_test.go
+++ b/pkg/reconciler/common/extensions_test.go
@@ -26,6 +26,14 @@ import (
 
 type TestExtension string
 
+func (t TestExtension) Manifests() []mf.Manifest {
+	manifest, err := mf.NewManifest("testdata/kodata/additional-manifests/additional-sa.yaml")
+	if err != nil {
+		return nil
+	}
+	return []mf.Manifest {manifest}
+}
+
 func (t TestExtension) Transformers(v1alpha1.KComponent) []mf.Transformer {
 	if t == "fail" {
 		return nil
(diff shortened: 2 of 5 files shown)


pkg/reconciler/common/extensions.go
```diff
@@ -39,6 +40,9 @@ func NoExtension(context.Context) Extension {
 
 type nilExtension struct{}
 
+func (nilExtension) Manifests() []mf.Manifest {
```
markusthoemmes, line 43: Should this take the KComponent to be able to dynamically decide based on Spec settings?

pkg/reconciler/common/extensions_test.go
```diff
@@ -26,6 +26,14 @@ import (
 
 type TestExtension string
 
+func (t TestExtension) Manifests() []mf.Manifest {
+	manifest, err := mf.NewManifest("testdata/kodata/additional-manifests/additional-sa.yaml")
+	if err != nil {
+		return nil
```
markusthoemmes, line 32: Feels like we should be able to return the error in case we want to abort.

pkg/reconciler/common/extensions.go
skonto: Yes good idea does not hurt anything to add it.

houshengbo:
https://github.com/knative/operator/pull/374
https://github.com/knative/operator/pull/379
@skonto 
I have added the support to specify the additional manifests.
Isn't it this PR a repetition?

skonto:
@houshengbo from a ux perspective and given the number of platform specific yaml files to maintain it is not. My scenario is not about exposing extra manifests which are accessed by some url etc. I dont want these files to be part of the crd contract. I have already mentioned why it is not in the description.

pkg/reconciler/common/extensions_test.go
```diff
@@ -26,6 +26,14 @@ import (
 
 type TestExtension string
 
+func (t TestExtension) Manifests(v1alpha1.KComponent) ([]mf.Manifest, error) {
+	manifest, err := mf.NewManifest("testdata/kodata/additional-manifests/additional-sa.yaml")
```
matzew, line 30: So in my custom extension, I can add new custom manifests by calling this. Only on code, right? not via Yaml?

Should we document that? Or at least capture this as a docs issue? Happy to help - I already have a use-case for this feature ;-)

skonto: @matzew there is already a way to add custom manifests via the CRD (yaml) check @houshengbo pointers above ;)
If you want to add the manifests via code without being exposed to the user (so they cannot be subject to illegal change) you can do it with this PR.

matzew:
> If you want to add the manifests via code without letting them being exposed to the user (so they cannot be subject to illegal change) you can do it with this PR. 

+1 that's exactly what I want!

/lgtm

leaving approval to others

houshengbo:
@skonto Are you leveraging the capability of the platform-specific extension?
Do you have an example of how to use this capability?
Knative lacks of docs on that aspect.

skonto:
@houshengbo yes check my link in the description, my plan is to create all the service monitor resources via a blueprint that is loaded from a local path and create the manifests dynamically.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -109,7 +109,11 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *v1alpha1.KnativeServ
 		common.CheckDeployments,
 		common.DeleteObsoleteResources(ctx, ks, r.installed),
 	}
-	manifest := r.manifest.Append()
+	platformManifests, err := r.extension.Manifests(ks)
```
houshengbo, line 112: @skonto plz update your local source code and do a rebase.
Your code is surely obsolete for now.

Another remark: as in the stages, we load the core manifests first, then the ingress manifests(you should have them after the rebase). And probably the extended manifests. It is order-critical. The one loaded first will be applied first, the one loaded in the end will overwrite the same ones before.
The question is if we want to make sure the extended manifest go over the core, we need to load them in the order: core first, ingress manifest second, and lastly the extended manifest.
But for the logic in this PR, it seems to be extended manifest first, core second... Considering changing the order.
You can create a function similar to `common.AppendTarget` for extension, and add it in the stage.

pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -106,7 +106,11 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ke *v1alpha1.KnativeEven
 		common.CheckDeployments,
 		common.DeleteObsoleteResources(ctx, ke, r.installed),
 	}
-	manifest := r.manifest.Append()
+	platformManifests, err := r.extension.Manifests(ke)
```
houshengbo, line 109: plz check my comment below for the serving. it applies here as well.

pkg/reconciler/knativeserving/knativeserving.go
houshengbo:
```
func (r *Reconciler) appendExtension(ctx context.Context, manifest *mf.Manifest, instance v1alpha1.KComponent) error {
	platformManifests, err := r.extension.Manifests(instance)
	if err != nil {
		return err
	}
	*manifest = manifest.Append(platformManifests)
	return nil
}
```

```
	stages := common.Stages{
		common.AppendTarget,
		ingress.AppendTargetIngresses,
		r.filterDisabledIngresses,
                r.appendExtension,
		r.transform,
		common.Install,
		common.CheckDeployments,
		common.DeleteObsoleteResources(ctx, ks, r.installed),
	}
```

skonto: thanks @houshengbo will take a look.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -108,6 +108,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *v1alpha1.KnativeServ
 		common.AppendTarget,
 		ingress.AppendTargetIngresses,
 		r.filterDisabledIngresses,
+		r.appendExtension,
```
matzew, line 111: should we name it `appendExtensionManifests` or the like?

We can do that in a different PR, but just wondering, since it's not appending the extension at whole, right ?

skonto: Ok .

matzew (approved):
/lgtm
/approve

PR #469, opened 2021-02-13
Move istio gateway's override setting into spec.ingress.istio

nak3:
This patch moves istio gateway's override setting into
spec.ingress.istio.

spec.knative-ingress-gateway and spec.cluster-local-gateway are
still valid but obsolete.

e.g. confgiure gateway override
```
kind: KnativeServing
metadata:
  name: ks
  namespace: knative-serving
spec:
  ingress:
    istio:
      enabled: true
      knative-local-gateway:
        selector:
          foo: bar
EOF
```

**Release Note**

```release-note
spec.knative-ingress-gateway and spec.cluster-local-gateway are deprecated.
Use `spec.ingres.istio.knative-ingress-gateway` and `spec.ingres.istio.knative-local-gateway` instead.
```


pkg/apis/operator/v1alpha1/knativeserving_types.go
```diff
@@ -112,6 +116,14 @@ type IngressConfigs struct {
 // IstioIngressConfiguration specifies options for the istio ingresses.
 type IstioIngressConfiguration struct {
 	Enabled bool `json:"enabled"`
+
+	// KnativeIngressGateway overrides the knative-local-gateway.
```
houshengbo, line 120: s/knative-local-gateway/knative-ingress-gateway

pkg/reconciler/knativeserving/ingress/istio_test.go
```diff
@@ -114,6 +119,30 @@ func TestGatewayTransform(t *testing.T) {
...
+			gatewayTransform(instance, log)(gateway)
+
+			got, ok, err = unstructured.NestedStringMap(gateway.Object, "spec", "selector")
+			util.AssertEqual(t, err, nil)
+			util.AssertEqual(t, ok, true)
+
+			if !cmp.Equal(got, tt.expected) {
+				t.Errorf("Got = %v, want: %v, diff:\n%s", got, tt.expected, cmp.Diff(got, tt.expected))
+			}
+
```
houshengbo, line 145: It may be necessary for us at this moment to run another test with both of spec.DeprecatedKnativeIngressGateway and spec.ingress.istio.KnativeIngressGateway, to see which one wins in terms of the value?

And with both of spec.DeprecatedClusterLocalGateway and spec.ingress.istio.KnativeLocalGateway, to see which one wins in terms of the value too?

maximilien (approved):
How long is the deprecation for? I would expect some type of time limit or better release limit, e.g., two major release from this one. Essentially, would like to avoid deprecation that last forever or are expected to last forever. We can avoid this by saying that this deprecation is for 2 releases and then remove that feature in 2 releases. Thoughts?

nak3:
> How long is the deprecation for?

The deprecation period should be same with Kubernetes's https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api.

> GA: 12 months or 3 releases (whichever is longer)
> Beta: 9 months or 3 releases (whichever is longer)
> Alpha: 0 releases


serving's field, for example [`imageDigest`](https://knative.dev/docs/reference/api/serving-api/), follow the same rule as:

> TODO(savitaashture) Remove deprecatedImageDigest. ref https://kubernetes.io/docs/reference/using-api/deprecation-policy for deprecation.

pkg/apis/operator/v1alpha1/knativeserving_types.go
nak3: Thank you. Done.

pkg/reconciler/knativeserving/ingress/istio_test.go
nak3: Sure, added it.

PR #470, opened 2021-02-13
Allow to configure Kourier gateway service-type

nak3:
When deploying Kourier, the gateway is always LoadBalancer.
It should be supported other service type such as `ClusterIP` and
`NodePort`.

e.g. You can configure `ClusterIP` with following CR.

```
kind: KnativeServing
metadata:
  name: ks
  namespace: knative-serving
spec:
  config:
    network:
      ingress.class: kourier.ingress.networking.knative.dev
  ingress:
    istio:
      enabled: false
    kourier:
      enabled: true
      service-type: "ClusterIP"
EOF
```

**Release Note**

```release-note
New configuration `spec.ingress.kourier.service-type` is available.
```


pkg/reconciler/knativeserving/ingress/kourier.go
```diff
@@ -18,53 +18,83 @@ package ingress
...
+			if err := scheme.Scheme.Convert(u, svc, nil); err != nil {
+				return err
+			}
+
+			serviceType := instance.Spec.Ingress.Kourier.ServiceType
+			switch serviceType {
+			case v1.ServiceTypeClusterIP, v1.ServiceTypeNodePort, v1.ServiceTypeLoadBalancer:
+				svc.Spec.Type = serviceType
+			default:
+				return fmt.Errorf("Unknown service type %q", serviceType)
```
markusthoemmes, line 91:
```suggestion
				return fmt.Errorf("unknown service type %q", serviceType)
```

pkg/reconciler/knativeserving/ingress/ingress.go
```diff
@@ -106,3 +106,10 @@ func AppendInstalledIngresses(ctx context.Context, manifest *mf.Manifest, instan
 	}
 	return getIngress(version, manifest)
 }
+
+func hasProviderLabel(u *unstructured.Unstructured) bool {
```
markusthoemmes, line 110: The need for this seems weird to me. Can we not apply the Transforms to **just** the respective Ingress' manifest and not run them later against the entire manifest?

nak3: @markusthoemmes Yes, it makes sense.
Now I tried to change it but the current filtering logic needs to be changed a lot and I think I should fix it in another PR...

houshengbo: @nak3 I do not think we need this function to check the label.
Check the function `ReconcileKind` in knativeserving.go:
```
	stages := common.Stages{
		common.AppendTarget,
		ingress.AppendTargetIngresses,
		r.filterDisabledIngresses,
		r.transform,
		common.Install,
		common.CheckDeployments,
		common.DeleteObsoleteResources(ctx, ks, r.installed),
	}
```

The r.filterDisabledIngresses(containing the ingress.Filters) runs before r.transform(containing the ingress.Transformers), so when we transform with `kourierTransformers`, the manifests have removed the other disabled ingresses. There is no need to check if it has the kourier label, because resources with other labels have been removed.

pkg/reconciler/knativeserving/ingress/ingress_test.go
```diff
@@ -447,6 +447,7 @@ func TestFilters(t *testing.T) {
 	}
 }
 
+// TODO: This test verifies the number of transformers. It should be rewritten by better test.
```
maximilien, line 450: I would elaborate on this in a issue and link here

pkg/reconciler/knativeserving/ingress/kourier_test.go
```diff
@@ -19,85 +19,116 @@ package ingress
...
-	}
-	result := &unstructured.Unstructured{}
-	err := scheme.Scheme.Convert(d, result, nil)
-	if err != nil {
-		t.Fatalf("Could not create unstructured Deployment: %v, err: %v", d, err)
+// removeProviderLabels removes labels. This util is used for tests without provider label.
+func removeLabels() mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		u.SetLabels(map[string]string{})
+		return nil
```
maximilien, line 132: I don’t understand why this always return nil. Why not make it a void function instead?

pkg/reconciler/knativeserving/ingress/kourier_test.go
```diff
@@ -19,85 +19,116 @@ package ingress
...
+				t.Fatalf("Failed to transform manifest: %v", err)
+			}
+
+			manifest, err = manifest.Transform(configureGWServiceType(tt.instance))
+			if err != nil {
+				t.Fatalf("Failed to transform manifest: %v", err)
+			}
+
+			for _, u := range manifest.Resources() {
+				if u.GetKind() == "Deployment" && u.GetName() == kourierControllerDeploymentName {
```
maximilien, line 103: Can we simplify this for loop content into one or better two functions that clearly indicate what’s going on. Especially since you have a nested for loop to deal with env. Hard to read and test IMO

maximilien (approved):
Left a few comments.

pkg/reconciler/knativeserving/ingress/kourier_test.go
nak3: The `removeLabels()` is used `manifest.Transform()` and the type is defined in:

https://github.com/knative/operator/blob/846d8762f2d1df4a0d080e30ff89c54cf845be02/vendor/github.com/manifestival/manifestival/transform.go#L13

And it needs to return `error `.

pkg/reconciler/knativeserving/ingress/kourier_test.go
nak3: Sure, done.

pkg/reconciler/knativeserving/ingress/ingress_test.go
nak3: Alright! Opened https://github.com/knative/operator/issues/477

pkg/reconciler/knativeserving/ingress/ingress.go
houshengbo: @markusthoemmes @nak3 
This is surely a good refactor, but it could take some effort.
Now we have the serving manifests and ingress manifests, with common transformers for both, serving transformers for serving and ingress transformers for ingress. It is also order-critical in the code to load which manifest first, and which transformer first. The current approach loads all manifests together, and run the transformers in the correct order. It lowered the performance, but simplified the logic. But we can work on that later.

Issue #442, opened 2021-01-26
Allow defining Pod Annotations from the KnativeServing Custom Resource 


rawc0der:
**Problem**
I have encountered a lot of use-cases where custom annotations need to be provided for the deployed resources. 
For my most recent context, the `KnativeServing` crd is lacking configuration options for injecting annotations into the `Activator pods` and making my life hard.

Example desired usage of custom annotation fields:
```
    apiVersion: operator.knative.dev/v1alpha1
    kind: KnativeServing
    metadata:
      name: knative-serving
      namespace: knative-serving
    spec:
      version: 0.XX.0
      config:
        defaults:
...   
       activator:
         podAnnotations: 
           custom/annotations: "values"
```

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)**
Which persona is this feature for?
- Event Consumers / Developers using or setting up Knative

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.
- Allow templating from helm and verify annotations are inserted. 

**Time Estimate (optional):**
How many developer-days do you think this may take to resolve?
 - 1

**Additional context (optional)**
Add any other context about the feature request here.
- If anyone thinks it would be an easy task for an outsider and has a few pointers, I am keen on contributing but am not sure where this fits right now. must catch up on reading the code base.
- I am sure this is not the only use case requiring users setting pod annotations, thus I propose to allow annotations on all the resources the CRD's are managing.

Thanks

Issue #382, opened 2020-12-08
Resource settings do not propagate


antoineco:
**Describe the bug**

Resources requirements defined as described at https://knative.dev/docs/install/operator/configuring-eventing-cr/#system-resource-settings do not propagate to the target deployment(s).

Tested in an already existing installation, didn't test in a new installation.

**Expected behavior**

Resources requirements do propagate.

**To Reproduce**

```
apiVersion: operator.knative.dev/v1alpha1
kind: KnativeEventing
metadata:
  name: knative-eventing
  namespace: knative-eventingf
spec:
  version: 0.19.2
  resources:
  - container: imc-dispatcher
    requests:
      cpu: 200m
      memory: 250Mi
    limits:
      cpu: "3"
      memory: 1200Mi
```

**Knative release version**

`v0.19.1`

**Additional context**

antoineco:
This worked, as recommended by @houshengbo: `container: dispatcher`.

Probably the documentation should be outdated, or the mapping should be somehow fixed.

Issue #337, opened 2020-10-29
Update HA logic


nak3:
**Describe the bug**

- As per current HA logic code in https://github.com/knative/operator/blob/18aca414e5109ee3f66b4431430616c65ef799ab/pkg/reconciler/knativeserving/common/ha.go 
it still checks `enabledComponents` in configmap. The parameter was removed by https://github.com/knative/serving/commit/584d41b1723f3a7aa7ab4181638793f47cb2a430.
- Also, the HA serving's components are basically "controller" "webhook" "autoscaler-hpa" "autoscaler" "domainmapping-webhook" but all of them are not covered.

**To Reproduce**

- One of the problem is that `high-availability.replicas` in KnativeServing does not scale up autoscaler.

houshengbo:
@nak3 Initially, the HA in operator only supported the leader-selected deployments:
https://knative.dev/docs/install/operator/configuring-serving-cr/#high-availability. It was just controller, autoscaler-hpa, and networking-istio.

Per enabledComponents, I think we probably need to keep it for a while, because operator maintain minor version minus 3. Old version of knative serving could still have it?

nak3:
Yeah, the operator maintain minor version minus 3, then `enabledComponents` should be remained.

For HA components, ingress plugins also needs to be considered once https://github.com/knative/operator/pull/312 was merged.

PR #472, opened 2021-02-15
Support spec.deployments to override configuration of system deployments

nak3:
Current spec has some problems:

- Annotations and Labels cannot be added safely into system pods.
- HA Logic is obsolete.
- HA setting applies to the all system pods. e.g. HA=2 means all deployments (webhook,controller,etc) scale to 2.
- HA setting does not work for Ingress manifests like Kourier and Contour.
- Resource setting does not work accurately if there are two pods have a same name container. e.g. webhook and istio-webhook have `webhook` container.
  ...etc.

To fix them, this patch introduces `deployments` field to override HA and Label/Annitations field for deployments.

Fixes https://github.com/knative/operator/issues/442 
~https://github.com/knative/operator/issues/382~ ~https://github.com/knative/operator/issues/337~

e.g. Override webhook and 3scale-kourier-controller

```
kind: KnativeServing
metadata:
  name: ks
  namespace: knative-serving
spec:
  high-availability:
    replicas: 1
  deployments:
  - name: webhook # Override webhook deployments.
    replicas: 2
    labels:
      foo: bar
  - name: 3scale-kourier-control
    replicas: 2
    labels:
      foo: bar
  ingress:
    istio:
      enabled: false
    kourier:
      enabled: true
  config:
    network:
      ingress.class: kourier.ingress.networking.knative.dev
EOF
```

**Release Note**

```release-note
Support `spec.deployments` to override configuration of system deployments such as label,annotation and replicas.
```


matzew:
Is there an issue captured for documenting these changes on `knative/docs` repo?

houshengbo:
@nak3 I would suggest NOT deprecate the spec.HA.
We can leave it as an option to configure all the deployments with the same number of replicas.
spec.deployment to overwrite if available.

nak3:
@matzew I haven't filed the change on `knative/docs` repo yet because I was (am still) not sure if this change will be accepted or not :worried:

@houshengbo Okay, I understand about spec.HA.
So, if we have the following configuration, `webhook` has `3` replicas, but other deployments have `2` replicas. Is that alright?

```
spec:
  high-availability:
    replicas: 2
  deployments:
  - name: webhook
    replicas: 3
```

For `spec.resources`, okay I will drop the change now.
But https://github.com/knative/operator/issues/382#issuecomment-741002226 needs to be solved somehow. istio-webhook and webhook already hit the issue.

houshengbo:
@nak3 Right, that's how we can plan the usage of the ha and the deployment replicas together.

pkg/reconciler/common/deployments_override.go
```diff
@@ -0,0 +1,91 @@
...
+	if deployment.Spec.Template.GetLabels() == nil {
+		deployment.Spec.Template.Labels = map[string]string{}
+	}
+	for key, val := range override.Labels {
+		deployment.Labels[key] = val
+		deployment.Spec.Template.Labels[key] = val
+	}
+}
+
+func replaceReplicas(u *unstructured.Unstructured, override *v1alpha1.DeploymentOverride) error {
```
houshengbo, line 86: Could you help to refactor this function into changing the deployment like replaceLabels?
Let's say we get rid of setting fields in the unstructured, but changing the fields in the deployment object.

nak3: Sure, updated. Thank you.

PR #480, opened 2021-02-17
Add HA support for autoscaler

nak3:
Since serving v0.19+, autoscaler supports HA[1] so operator also should
support it.

[1] https://github.com/knative/serving/releases/tag/v0.19.0
> Autoscaler now supports multiple pods. Autoscaler Deployment needs to be scaled to 0 first then scaled to other replica value.

**Release Note**

```release-note
Support HA for serving's autoscaler.
```

/cc @houshengbo @maximilien @markusthoemmes

### DIFF
--- a/pkg/reconciler/common/ha.go
+++ b/pkg/reconciler/common/ha.go
@@ -19,6 +19,7 @@ package common
 import (
 	mf "github.com/manifestival/manifestival"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


maximilien (approved):
How hard is it to add a e2e or any test for this?

nak3:
e2e for HA should be added. Opened https://github.com/knative/operator/issues/482.

PR #485, opened 2021-02-22
Fetcher: support ingress alternatives, fetch from GCS

evankanderson:
## Proposed Changes

* Add support for fetch from authenticated S3/GCS/Azure buckets, and unauthenticated GCS buckets (which works for the public release)
* Add support for an `alternatives` package to support the current `ingress` packages.
* Fix `yaml` tags that should now be `json` tags.

**Release Note**

```release-note
Added automation support to fetch release assets from blob storage services.
```


evankanderson:
/assign @houshengbo 

Note that this renames a few `kodata` files that were moved with the `ingress` changes, and restores some misspellings in 0.17 eventing manifests.

This should also enable the nightly codegen to run, as it will no longer need a `$GITHUB_TOKEN` to proceed in the GCS case.

houshengbo:
Nice!!
/lgtm
/approve

PR #492, opened 2021-02-25
Download the ingress net-istio into the correct directory

houshengbo:
## Proposed Changes

* The prow downloads the latest net-istio ingress into the correct directory under kodata.


maximilien (approved):
/lgtm

PR #498, opened 2021-03-03
Change container name of the sugar-controller to support separate resource requests

Fabian-K:
## Proposed Changes
This PR changes the container name of the sugar-controller Deployment from `controller` to `sugar-controller`. This enables to set separate resource requests using https://github.com/knative/operator/blob/master/docs/configuration.md#specresources. Currently, the sugar-controller and the imc-controller both share the container name `controller`

**Release Note**

```release-note
The container name of the sugar-controller Deployment changes from `controller` to `sugar-controller`
```

### DIFF
--- a/cmd/operator/kodata/knative-eventing/0.18.0/6-eventing-sugar-controller.yaml
+++ b/cmd/operator/kodata/knative-eventing/0.18.0/6-eventing-sugar-controller.yaml
@@ -30,7 +30,7 @@ spec:
     spec:
       serviceAccountName: eventing-controller
       containers:
-      - name: controller
+      - name: sugar-controller
         image: gcr.io/knative-releases/knative.dev/eventing/cmd/sugar_controller@sha256:c1836e244b4e31c130e72de785bf7ff0d2d481ba7bef08626a7d557cd6a8dfc1
         env:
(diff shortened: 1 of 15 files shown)


Fabian-K:
Oh is this the wrong approach and this should be changed somewhere else?

Fabian-K:
Ok, I *think* this needs to be changed within kantive eventing (https://github.com/knative/eventing/blob/master/config/sugar/500-controller.yaml#L33) and once it´s merged there and part of a release, this will automatically be available in the operator?

houshengbo:
@Fabian-K The manifests are in sync with the released artifacts available at https://github.com/knative/eventing/releases
We usually do not change them, unless there is automated sync-up.

Fabian-K:
Thanks for the reply! I don´t think that there is the need to change existing releases, would be fine for me if this only applies to future releases. I´ll probably open a PR on the knative eventing repo.

Targeting deployments is also an option, however, this might break if there are at some point multiple containers within a deployment. Not sure though if that's a realistic concern ;)

maximilien (commented):
/ok-to-test

houshengbo:
close this PR, since we do not modify the manifests bundled in the operator.

PR #500, opened 2021-03-04
Refactor the test cases of checking the resources

houshengbo:
## Proposed Changes

* This PR adds the test functions able to compare the actual manifest resources with the expected manifest resources.

### DIFF
--- a/pkg/reconciler/common/testing/util.go
+++ b/pkg/reconciler/common/testing/util.go
@@ -19,6 +19,10 @@ import (
 	"reflect"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/api/equality"
+
+	mf "github.com/manifestival/manifestival"
+
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/knativeserving/ingress/ingress_test.go
```diff
@@ -78,8 +77,9 @@ func TestAppendInstalledIngresses(t *testing.T) {
 	tests := []struct {
 		name                 string
 		instance             servingv1alpha1.KnativeServing
-		expected             bool
+		expectedNoErr        bool
```
nak3, line 80: Do we need this `expectedNoErr`? The test cases always expect `true`.

pkg/reconciler/knativeserving/ingress/ingress_test.go
```diff
@@ -78,8 +77,9 @@ func TestAppendInstalledIngresses(t *testing.T) {
...
 		expectedResourcesNum int
```
nak3, line 81: `expectedResourcesNum` is not used. Let's remove it.

pkg/reconciler/knativeserving/ingress/ingress_test.go
```diff
@@ -21,52 +21,51 @@ import (
...
-		expectedResourcesNum: 0,
+		name:          "Unavailable ingresses",
+		version:       "0.16.1",
+		expectedNoErr: false,
 	}, {
-		name:                 "Missing version",
-		targetVersion:        "",
-		expected:             true,
-		expectedResourcesNum: 0,
+		name:                "Missing version",
```
nak3, line 53: When version was missing, no error and no expectedIngressPath. Is that correct? (I just wanted to confirm it with you.)

pkg/reconciler/knativeserving/ingress/ingress_test.go
```diff
@@ -21,52 +21,51 @@ import (
...
 	}}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			manifest, _ := mf.ManifestFrom(mf.Slice{})
-			err := getIngress(tt.targetVersion, &manifest)
-			util.AssertEqual(t, err == nil, tt.expected)
-			util.AssertEqual(t, len(manifest.Resources()), tt.expectedResourcesNum)
+			err := getIngress(tt.version, &manifest)
+			util.AssertEqual(t, err == nil, tt.expectedNoErr)
```
nak3, line 63: Can we check the `os.IsNotExist(err)` accurately instead of `err == nil`?

pkg/reconciler/knativeserving/ingress/ingress_test.go
```diff
@@ -21,52 +21,51 @@ import (
...
 	defer os.Unsetenv(common.KoEnvKey)
 
 	tests := []struct {
-		name                 string
-		targetVersion        string
-		expected             bool
-		expectedResourcesNum int
+		name                string
+		version             string
+		expectedNoErr       bool
```
nak3, line 41: I personally like `expectError` instead of `expectedNoErr`. Because it is simple and we can omit `false` as it is default.

pkg/reconciler/common/testing/util.go
```diff
@@ -19,6 +19,10 @@ import (
 	"reflect"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/api/equality"
+
```
nak3, line 23: Can you remove blank lines around here?

pkg/reconciler/common/testing/util.go
```diff
@@ -85,3 +89,92 @@ func AssertDeepEqual(t *testing.T, actual, expected interface{}) {
 	}
 	t.Fatalf("expected does not deep equal actual. \nExpected: %T %+v\nActual:   %T %+v", expected, expected, actual, actual)
 }
+
+func ResourceMatchWithPath(actual mf.Manifest, expectedManifestPath string) bool {
```
nak3, line 93: Is this function called used somewhere?

pkg/reconciler/common/testing/util.go
```diff
@@ -85,3 +89,92 @@ func AssertDeepEqual(t *testing.T, actual, expected interface{}) {
...
+	expected, err := mf.NewManifest(expectedManifestPath)
+	if err != nil {
+		return false
+	}
+	return ResourceDeepMatch(actual, expected)
+}
+
+// ResourceDeepMatch returns true if the resources in the actual manifest match exactly the same resources in
+// the expected manifest.
+func ResourceDeepMatch(actual, expected mf.Manifest) bool {
```
nak3, line 130: Can use lower case e.g. `resourceDeepMatch()` when the function can be private?

pkg/reconciler/knativeserving/ingress/ingress_test.go
houshengbo: The func getIngress with the missing version returns empty manifest and nil as error.

pkg/reconciler/knativeserving/ingress/ingress_test.go
houshengbo: will do

pkg/reconciler/common/testing/util.go
houshengbo: will do

pkg/reconciler/knativeserving/ingress/ingress_test.go
houshengbo: I will change the `expectError` into a real error, instead of the bool.
If it is nil, check the equality. if it is an error, check the string of the error.

pkg/reconciler/knativeserving/ingress/ingress_test.go
houshengbo: I will change the `expectError` into a real error, instead of the bool.
If it is nil, check the equality. if it is an error, check the string of the error.

pkg/reconciler/knativeserving/ingress/ingress_test.go
houshengbo: I will change the `expectError` into a real error, instead of the bool.
If it is nil, check the equality. if it is an error, check the string of the error.

pkg/reconciler/knativeserving/ingress/ingress_test.go
```diff
@@ -100,16 +101,16 @@ func TestAppendInstalledIngresses(t *testing.T) {
 			},
 			Status: servingv1alpha1.KnativeServingStatus{},
 		},
-		expected:             true,
-		expectedResourcesNum: numberIngressResource,
+		expectedIngressPath: os.Getenv(common.KoEnvKey) + "/ingress/0.18",
+		expectedErr:         nil,
 	}}
```
nak3, line 106: Do we want to add one more test case `"Unavailable ingresses"` with invalid version?

houshengbo: done

PR #509, opened 2021-03-08
Filter the redundant resources in the target manifest

### LINKED ISSUE #508: Overwriting existing deployment resources spec.additionalManifests ends up with endless reconciling
**Describe the bug**
Leveraging spec.additionalManifests, we are able to specify the customized resources based on the default one.
If the resources in spec.additionalManifests already exist in default manifests, they should replace the default ones.
However, coming the deployment resource, the reconciling loop will keep on reconciling for ever, since the deployment will run into running-> terminating -> running... for ever.

**Expected behavior**
Reconciling loop should succeed ultimately.

**To Reproduce**
The spec.additionalManifests specifies the URL with existing deployment resources in the default manifests.

**Knative release version**
ANY supporting spec.additionalManifests.

houshengbo:
Fixes #508

### DIFF
--- a/pkg/reconciler/common/releases.go
+++ b/pkg/reconciler/common/releases.go
@@ -65,19 +65,17 @@ func TargetVersion(instance v1alpha1.KComponent) string {
 // TargetManifest returns the default manifest for the TargetVersion or the manifest for the TargetVersion specified
 // with spec.manifests
 func TargetManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
-	version := TargetVersion(instance)
-	manifestsPath := targetManifestPath(version, instance)
-	return getManifestWithVersionValidation(version, manifestsPath, instance)
+	manifestsPath := targetManifestPath(instance)
+	return getManifestWithVersionValidation(manifestsPath, instance)
 }
 
 // TargetAdditionalManifest returns the manifest for the TargetVersion specified with spec.additionalManifests.
 func TargetAdditionalManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
-	version := TargetVersion(instance)
-	additionalManifestsPath := additionalManifestPath(version, instance)
+	additionalManifestsPath := additionalManifestPath(instance)
 	if additionalManifestsPath == "" {
 		return mf.Manifest{}, nil
 	}
-	return getManifestWithVersionValidation(version, additionalManifestsPath, instance)
+	return getManifestWithVersionValidation(additionalManifestsPath, instance)
 }
 
 // InstalledManifest returns the version currently installed, which is
@@ -155,7 +153,8 @@ func getVersionKey(instance v1alpha1.KComponent) string {
 	return ""
 }
 
-func getManifestWithVersionValidation(version, manifestsPath string, instance v1alpha1.KComponent) (mf.Manifest, error) {
+func getManifestWithVersionValidation(manifestsPath string, instance v1alpha1.KComponent) (mf.Manifest, error) {
+	version := TargetVersion(instance)
 	manifests, err := FetchManifest(manifestsPath)
 	if err != nil {
 		if len(instance.GetSpec().GetManifests()) == 0 && len(instance.GetSpec().GetAdditionalManifests()) == 0 {
@@ -225,18 +224,19 @@ func componentDir(instance v1alpha1.KComponent) string {
 	return ""
 }
 
-func additionalManifestPath(version string, instance v1alpha1.KComponent) string {
+func additionalManifestPath(instance v1alpha1.KComponent) string {
 	// Create the comma-separated string for URLs in spec.additionalManifests
 	addManifests := instance.GetSpec().GetAdditionalManifests()
 	urls := make([]string, 0, len(addManifests))
 	for _, manifest := range addManifests {
-		url := strings.ReplaceAll(manifest.Url, VersionVariable, version)
+		url := strings.ReplaceAll(manifest.Url, VersionVariable, TargetVersion(instance))
 		urls = append(urls, url)
 	}
 	return strings.Join(urls, COMMA)
 }
 
-func targetManifestPath(version string, instance v1alpha1.KComponent) string {
+func targetManifestPath(instance v1alpha1.KComponent) string {
+	version := TargetVersion(instance)
 	manifests := instance.GetSpec().GetManifests()
 	// Create the comma-separated string as the URL to retrieve the manifest
 	urls := make([]string, 0, len(manifests))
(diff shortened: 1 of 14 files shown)


pkg/reconciler/common/releases.go
```diff
@@ -73,9 +89,13 @@ func TargetManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
 // So we return the target manifest if status.version is empty.
 func InstalledManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
 	current := instance.GetStatus().GetVersion()
+	fmt.Println("the current is ")
```
maximilien, line 92: Is this DEBUG statements left by mistake... I know it’s WIP

maximilien (commented):
/ok-to-test

I’ll wait for WIP removal to do more thorough review.

pkg/reconciler/common/releases.go
houshengbo: will remove it.

pkg/reconciler/common/releases.go
```diff
@@ -230,26 +253,29 @@ func targetManifestPath(version string, instance v1alpha1.KComponent) string {
...
+		additionalManifestPath := additionalManifestPath(version, instance)
+		pathArrayPath = strings.Join([]string{pathArrayPath, additionalManifestPath}, COMMA)
 
+	} else if len(instance.GetSpec().GetManifests()) > 0 {
+		// If spec.manifests is not empty, we leverage status.manifests
+		// to save the complete manifest path.
+		pathArrayPath = targetManifestPath(TargetVersion(instance), instance)
+	}
+	if pathArrayPath != "" {
+		return strings.Split(pathArrayPath, COMMA)
```
nak3, line 275: Why do we need to call `strings.Join` above and `strings.Split` here? 
Can `pathArrayPath` be string array?

pkg/reconciler/common/releases.go
```diff
@@ -62,9 +62,22 @@ func TargetVersion(instance v1alpha1.KComponent) string {
...
+}
+
+// TargetAdditionalManifest returns the manifest for the TargetVersion specified with spec.additionalManifests.
+func TargetAdditionalManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
+	version := TargetVersion(instance)
+	additionalManifestsPath := additionalManifestPath(version, instance)
+	if additionalManifestsPath == "" {
+		return mf.Manifest{}, nil
+	}
+	return getManifestWithVersionValidation(version, additionalManifestsPath, instance)
```
nak3, line 80: `additionalManifestPath()` could return comma separated manifests. If `additionalManifestsPath` contains comma, `getManifestWithVersionValidation()` could work properly?

pkg/reconciler/common/releases_test.go
```diff
@@ -731,59 +859,94 @@ func TestTargetManifest(t *testing.T) {
 				},
 			},
 		},
-		expectedNumResources: 2,
-		expectedError:        nil,
+		expectedManifestsPath: "",
+	}}
```
nak3, line 863: Can you add a test case which verifies that multiple manifest paths in `AdditionalManifests`?

pkg/reconciler/common/releases.go
```diff
@@ -213,6 +225,17 @@ func componentDir(instance v1alpha1.KComponent) string {
 	return ""
 }
 
+func additionalManifestPath(version string, instance v1alpha1.KComponent) string {
```
nak3, line 228:
```suggestion
func additionalManifestPath(instance v1alpha1.KComponent) string {
```

Is it not alright without version arg? `version := TargetVersion(instance)`  should be called inside the func.
(The same commend for `targetManifestPath`.)

pkg/reconciler/common/releases.go
houshengbo: Yes, if there are multiple paths, `getManifestWithVersionValidation` works fine with parsing and reading them, as long as they are concatenated with comma.

pkg/reconciler/common/releases.go
houshengbo: done

pkg/reconciler/common/releases.go
```diff
@@ -230,26 +253,29 @@ func targetManifestPath(version string, instance v1alpha1.KComponent) string {
...
+		// If spec.additionalManifests is not empty, we leverage status.manifests
 		// to save the complete manifest path.
-		return strings.Split(targetManifestPath(TargetVersion(instance), instance), COMMA)
-	}
+		version := TargetVersion(instance)
+		pathArrayPath = targetManifestPath(version, instance)
+		additionalManifestPath := additionalManifestPath(version, instance)
+		pathArrayPath = strings.Join([]string{pathArrayPath, additionalManifestPath}, COMMA)
 
+	} else if len(instance.GetSpec().GetManifests()) > 0 {
```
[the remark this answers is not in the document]
houshengbo: I am thinking of whether it is a good idea to ALWAYS return the path for status.manifests, no matter it is local default path, or spec.manifests, or the combination of local path and spec.additionalManifests, or even the combination of spec.manifests and spec.additionalManifests.

So far, if neither spec.additionalManifests or spec.Manifests are available, we leave it blank.

pkg/reconciler/common/releases.go
houshengbo: pathArrayPath itself could be a string with multiple paths, separated with comma. additionalManifestPath too.
We connect them together, and split with comma, it will end up with an array of all available paths as each item. NOT JUST pathArrayPath and additionalManifestPath, two items.

pkg/reconciler/common/releases_test.go
houshengbo: done

pkg/reconciler/common/releases.go
```diff
@@ -230,27 +253,17 @@ func targetManifestPath(version string, instance v1alpha1.KComponent) string {
...
-	return strings.Join(urls, COMMA)
+	return manifestPath
 }
 
 func targetManifestPathArray(instance v1alpha1.KComponent) []string {
-	if len(instance.GetSpec().GetManifests()) > 0 || len(instance.GetSpec().GetAdditionalManifests()) > 0 {
-		// If either spec.manifests or spec.additionalManifests is not empty, we leverage status.manifests
-		// to save the complete manifest path.
-		return strings.Split(targetManifestPath(TargetVersion(instance), instance), COMMA)
+	pathArrayPath := targetManifestPath(instance)
```
nak3, line 260: `pathArrayPath` sounds strange. Also `pathArrayPath` does not have array but comma separated strings, correct?

pkg/reconciler/common/stages_test.go
```diff
@@ -30,17 +30,184 @@ import (
...
-	util.AssertEqual(t, len(manifest.Resources()), 0)
-	stages.Execute(context.TODO(), &manifest, &v1alpha1.KnativeServing{Spec: v1alpha1.KnativeServingSpec{
-		CommonSpec: v1alpha1.CommonSpec{
-			Version: "0.16.0",
+
+	tests := []struct {
+		name                   string
+		component              v1alpha1.KComponent
+		expectedManifestsPath  string
+		expectedContainingPath string
```
nak3, line 41: `expectedContainingPath` is not used in the test.

pkg/reconciler/common/stages_test.go
```diff
@@ -30,17 +30,184 @@ import (
...
+
+func TestStagesExecuteInstalledManifests(t *testing.T) {
+	koPath := "testdata/kodata"
+	os.Setenv(KoEnvKey, koPath)
+	defer os.Unsetenv(KoEnvKey)
+
+	testRepetition := []struct {
+		name                      string
+		component                 v1alpha1.KComponent
+		expectedContainingPath    string
```
nak3, line 139: Same with above. `expectedContainingPath` and `expectedNotContainingPath` are not used.

pkg/reconciler/common/releases.go
houshengbo: I changed it into target path.

pkg/reconciler/common/stages_test.go
houshengbo: removed.

pkg/reconciler/common/stages_test.go
houshengbo: removed.

PR #532, opened 2021-03-25
Refactor the cache mechanism

### LINKED ISSUE #242: Refactor the cache mechanism
After we implement the support of customized yamls, there are several work items related to the cache mechanism:

1. If multiple customized yamls are specified, but the order is different, how can we save it as the same record.
2. If the links of customized yamls remain the same, are we able to force a refresh of the cache? It is possible that the yaml contents linked by the URLs have been changed.

houshengbo:
Fixes #242

### DIFF
--- a/pkg/reconciler/common/releases.go
+++ b/pkg/reconciler/common/releases.go
@@ -213,6 +213,11 @@ func FetchManifest(path string) (mf.Manifest, error) {
 	return result, err
 }
 
+// ClearCache removes all the records saved in the cache.
+func ClearCache() {
+	cache = map[string]mf.Manifest{}
+}
+
 func componentDir(instance v1alpha1.KComponent) string {
 	koDataDir := os.Getenv(KoEnvKey)
 	switch instance.(type) {
--- a/pkg/reconciler/common/releases.go
+++ b/pkg/reconciler/common/releases.go
@@ -295,8 +295,8 @@ func targetManifestPath(instance v1alpha1.KComponent) string {
 }
 
 func targetManifestPathArray(instance v1alpha1.KComponent) []string {
-	manifestPaths := make([]string, 0)
 	targetMPath := targetManifestPath(instance)
+	manifestPaths := []string{targetMPath}
 	manifestPaths = append(manifestPaths, targetMPath)
 	if len(instance.GetSpec().GetAdditionalManifests()) > 0 {
 		// If spec.additionalManifests is not empty, we append it to the target path.
@@ -311,13 +311,11 @@ func installedManifestPath(version string, instance v1alpha1.KComponent) []strin
 		return manifests
 	}
 
-	manifestPaths := make([]string, 0)
 	localPath := filepath.Join(componentDir(instance), version)
 	if _, err := os.Stat(localPath); !os.IsNotExist(err) {
-		manifestPaths = append(manifestPaths, localPath)
+		return []string{localPath}
 	}
-
-	return manifestPaths
+	return []string{}
 }
 
 // SanitizeSemver always adds `v` in front of the version.
--- a/pkg/reconciler/common/releases.go
+++ b/pkg/reconciler/common/releases.go
@@ -66,7 +66,10 @@ func TargetVersion(instance v1alpha1.KComponent) string {
 // with spec.manifests
 func TargetManifest(instance v1alpha1.KComponent) (mf.Manifest, error) {
 	manifestsPath := targetManifestPath(instance)
-	return getManifestWithVersionValidation(manifestsPath, instance)
+	if len(instance.GetSpec().GetManifests()) == 0 {
+		getManifestWithVersionValidation(manifestsPath, instance, FetchManifest)
+	}
+	return getManifestWithVersionValidation(manifestsPath, instance, FetchManifestNoCache)
 }
 
 // TargetAdditionalManifest returns the manifest for the TargetVersion specified with spec.additionalManifests.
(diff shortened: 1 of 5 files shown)


houshengbo:
cc @Cynocracy

pkg/reconciler/common/releases_test.go
```diff
@@ -980,3 +988,96 @@ func TestTargetManifestPathArray(t *testing.T) {
...
+			Status: v1alpha1.KnativeServingStatus{
+				Version: "0.12.0",
+			},
+		},
+		expectedManifestsPath: "testdata/kodata/empty/empty-resource.yaml",
+	}}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			m, _ := InstalledManifest(test.component)
```
nak3, line 1063:
```suggestion
                        m, err := InstalledManifest(test.component)
                        util.AssertEqual(t, err, nil)
```

I think we should verify that error is nil here.

pkg/reconciler/common/releases_test.go
```diff
@@ -980,3 +988,96 @@ func TestTargetManifestPathArray(t *testing.T) {
...
+				Manifests: []string{"testdata/kodata/knative-serving/0.16.1/serving-crd.yaml" + "," +
+					"testdata/kodata/knative-serving/0.16.1/serving-core.yaml",
+					"testdata/kodata/additional-manifests/additional-resource.yaml"},
+			},
+		},
+		expectedManifestsPath: "testdata/kodata/knative-serving/0.16.1/serving-crd.yaml" + "," +
+			"testdata/kodata/knative-serving/0.16.1/serving-core.yaml" + "," +
+			"testdata/kodata/additional-manifests/additional-resource.yaml",
+	}, {
+		name: "knative-serving with status.version unavailable in kodata",
```
nak3, line 1047: Out of curious, this case (Even if we set spec.version to 0.16.1, status becomes 0.12.0) does happen?

houshengbo: It is possible.
We have 0.12.0 knative serving installed. After a long time, users installed the latest version of the operator. They changed the spec.version into 0.16.1, and apply the CR. The operator will failed the validation by providing the error message in the CR, that unable to migrate across multiple minor versions.

pkg/reconciler/common/releases.go
```diff
@@ -213,6 +237,20 @@ func FetchManifest(path string) (mf.Manifest, error) {
 	return result, err
 }
 
+// FetchManifestNoCache returns the manifest by reading them from the path, and saves them in the cache.
+func FetchManifestNoCache(path string) (mf.Manifest, error) {
+	result, err := mf.NewManifest(path)
+	if err == nil {
+		cache[path] = result
```
maximilien, line 244: So if the cache already had results you will overwrite and not return from cache?

maximilien: Or maybe you are checking that somewhere else?

pkg/reconciler/common/releases_test.go
```diff
@@ -980,3 +988,97 @@ func TestTargetManifestPathArray(t *testing.T) {
...
+		})
+	}
+}
+
+func TestCache(t *testing.T) {
+	// Make sure to start with empty cache
+	ClearCache()
+	util.AssertEqual(t, len(cache), 0)
+	expectedPath := "testdata/kodata/knative-serving/0.16.1/"
+	manifest, _ := mf.NewManifest(expectedPath)
```
maximilien, line 1077: Why not assert the error here and fail?

pkg/reconciler/common/releases_test.go
```diff
@@ -980,3 +988,97 @@ func TestTargetManifestPathArray(t *testing.T) {
...
+			Status: v1alpha1.KnativeServingStatus{
+				Version: "0.12.0",
+			},
+		},
+		expectedManifestsPath: "testdata/kodata/empty/empty-resource.yaml",
+	}}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			m, err := InstalledManifest(test.component)
```
maximilien, line 1063: Fail on `err` here?

maximilien (changes requested):
Left a few comments. Of course so long tests are passing this should be OK since in theory no externally visible behavior should creep out ... :)

pkg/reconciler/common/releases.go
houshengbo: This function always reads from the path, but save it into the cache.
Operator needs to know what has been installed in the cluster. This is why we still save it into the cache.

pkg/reconciler/common/releases_test.go
houshengbo: As I add the comment on Line 1064, this function should not raise error.
If the manifest is unavailable, return manifest with no resource.

pkg/reconciler/common/releases.go
```diff
@@ -213,6 +237,20 @@ func FetchManifest(path string) (mf.Manifest, error) {
 	return result, err
 }
 
+// FetchManifestFromPath returns the manifest by reading them from the path, and saves them in the cache.
+func FetchManifestFromPath(path string) (mf.Manifest, error) {
```
nak3, line 241:
```suggestion
func fetchManifestFromPath(path string) (mf.Manifest, error) {
```

Can you make this function private? It is not called from other packages.

nak3 (commented):
LGTM except for one nit.

PR #545, opened 2021-04-06
Create the PR to prepare the 0.22 release

houshengbo:
## Proposed Changes

* This reason for this PR is that 0.18 is about to remove as 0.22 is about to release, since we maintain 4 minor versions.
* This change makes sure that we use the available ingress versions to run the test cases.
* This PR updates all the necessary dependencies for knative serving and eventing at 0.22.


evankanderson:
Hmm, it looks like we only needed two manual changes:
* test/e2e-common.sh
* pkg/reconciler/knativeserving/ingress/ingress_test.go

I'm wondering if we can get this down to an automated PR so that a release lead can use a releasability check and then cut the release without needing much specific expertise.

Issue #565, opened 2021-04-28
Knative operator looks for "istio-system" namespace when custom istio gateway is set


jakoberpf:
**Describe the bug**
When we deploy the knative operator and set the custom `knative-ingress-gateway` to another gateway and namespace, the knative serving controller still looks for the `istio-system` namespace and fails to get ready.

operator log: `failed to apply non rbac manifest: namespaces \"istio-system\" not found`

describe of KnativeServing: `Install failed with message: namespaces "istio-system" not found`

**Expected behavior**
When setting the custom gateway, knative doesn't look for the gateway in `istio-system`, but the defined namespace for example `servicemesh`.

**To Reproduce**
Deploy istio in custom namespace and knative with the following configs.

**Knative release version**
Knative Operator: v0.21.0 - v0.22.0

**Additional context**
Serving Configuration:

```---
apiVersion: operator.knative.dev/v1alpha1
kind: KnativeServing
metadata:
  name: platform-knative-serving
spec:
  version: 0.22.0
  config:
    domain:
    istio:
      gateway.knative-serving.knative-ingress-gateway: "app-gateway.servicemesh.svc.cluster.local"
  ingress:
    istio:
      enabled: true
      knative-ingress-gateway:
        selector:
          istio: custom-gateway
```

If we have an obviously misconfiguration please point me out...

nak3:
I think it is bug in https://github.com/knative/operator/blob/main/pkg/reconciler/knativeserving/common/ingress_service.go#L28.
Perhaps creating a `istio-system` namespace could be a workaround? (operator will deploy `knative-local-gateway` svc though.)

jakoberpf:
@nak3 Creating the istio-system namespace created removed the error and created the `knative-local-gateway` as you predicted. The code snipped tells me that the `istio-system` namespace is hard coded and therefore can be changed?

Furthermore when I read the [documentation](https://github.com/knative/operator/blob/main/docs/configuration.md#speccluster-local-gateway) right, we can not set the operator to create this svc in another namespace but would have to create our own local gateway first and then give it the operator?

houshengbo:
@nak3 @jakoberpf 
The line https://github.com/knative/operator/blob/main/pkg/reconciler/knativeserving/common/ingress_service.go#L28 may NOT be the reason leading to the issue, because this line only makes sure the `Service` named `knative-local-gateway` is installed under the default ns istio-system.

jakoberpf:
@houshengbo 
[...]

PR #567, opened 2021-04-29
Clear OwnerReferences on knative-local-gateway Service

### LINKED ISSUE #507: Knative-local-gateway gets deleted and never re-created
**Describe the bug**
We had been using Knative 0.17.0 without much issue for the last few months. We wanted to upgrade to something a little newer and we've hit a bit of a wall getting things working because of the changes surrounding the `cluster-local-gateway` to `knative-local-gateway` name change. Effectively after upgrade, we lose the ability to communicate with our isitio ingress gateway and also within the cluster. Our knative services don't get registered in the cluster DNS and so we can't resolve them. E.g. we can call something like `files-00001.svc.cluster.local` but not `files.svc.cluster.local`.

We think that the issue is related to the serving reconciler. I can see in the logs for the operator that it tries 4 times to create the knative-local-gateway service in `istio-system`. It never logs any errors but also never manages to create the knative-local-gateway service.

**Questions**
* Is there any debug logging i can turn on in the operator to let me know if/why this process has failed?
* If there is some funny RBAC issue going on here - is there anywhere I can look to find that out?
* Is there any way to make the reconciler not delete the service `knative-local-gateway` in `istio-system` e.g. can I set some parameters to make it match what it expects?

**Expected behavior**
After deploy there should be a service called `knative-local-gateway` in the `istio-system` namespace. Knative services should be available in DNS within the cluster.

**To Reproduce**
1. Prepare a 0.17.0 installation with an `istio-minimal-operator.yml` file with the following contents:
```
---
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
  name: istio-controlplane
spec:
  values:
    global:
      proxy:
        autoInject: disabled
      useMCP: false
      # The third-party-jwt is not enabled on all k8s.
      # See: https://istio.io/docs/ops/best-practices/security/#configure-third-party-service-account-tokens
      jwtPolicy: first-party-jwt

  addonComponents:
    pilot:
      enabled: true
    prometheus:
      enabled: false

  components:
    ingressGateways:
      - name: istio-ingressgateway
        enabled: true
      - name: cluster-local-gateway
        enabled: true
        label:
          istio: cluster-local-gateway
          app: cluster-local-gateway
        k8s:
          service:
            type: ClusterIP
            ports:
              - port: 15020
                name: status-port
              - port: 80
                targetPort: 8080
                name: http2
              - port: 443
                targetPort: 8443
                name: https
```
and a `knative-serving-operator.yml` file with the following:
```
apiVersion: v1
kind: Namespace
metadata:
 name: knative-serving
---
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  version: {{ .Values.knative.version }}
  config:
    {{- if eq .Values.cloud "local" }}
    deployment:
      # List of repositories for which tag to digest resolving should be skipped
      # This list must include any registries that are used by knative and
      # to use the  local registry mirror - it appears that knative ignores
      # insecure-registry settings and /etc/hosts
      registriesSkippingTagResolving: |-
        docker.pkg.github.com,gcr.io,k8s.gcr.io,mcr.microsoft.com,docker.io,registry-1.docker.io
    {{- end }}
    # Set all knative services to be internal only. An API Gateway is required
    # to enable access to any Knative service. If this is changed to a public
    # domain name, existing services must be labeled with:
    # serving.knative.dev/visibility: cluster-local
    # (as of knative v0.20: https://knative.dev/v0.20-docs/serving/cluster-local-route/)
    domain:
      svc.cluster.local: ""
```
2. Upgrade the operator to 0.21.0
3. Update the `istio-minimal-operator.yml` to have the following contents:
```
---
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
  name: istio-controlplane
spec:
  values:
    global:
      proxy:
        autoInject: disabled
        logLevel: debug
      useMCP: false
      # The third-party-jwt is not enabled on all k8s.
      # See: https://istio.io/docs/ops/best-practices/security/#configure-third-party-service-account-tokens
      jwtPolicy: first-party-jwt

  addonComponents:
    pilot:
      enabled: true

  components:
    ingressGateways:
      - name: istio-ingressgateway
        enabled: true
      - name: knative-local-gateway
        enabled: true
        label:
          serving.knative.dev/release: "v0.21.0"
          networking.knative.dev/ingress-provider: istio
        k8s:
          service:
            type: ClusterIP
            selector:
              istio: ingressgateway
            ports:
              - name: http2
                port: 80
                targetPort: 8081
```
and the `knative-serving-operator` file remains the same.

**Knative release version**
Knative 0.19.0, 0.20.0, 0.21.0
istio versions: 1.8.3, also tried 1.9.1

polivbr:
Kubernetes v1.20+ does not allow cross-namespace owner references
and will cause the resource to be considered invalid and will
garbage collect it.

Because the knative-local-gateway service has its namespace
changed to istio-system, it's owner reference to the operator
is invalid, and it must be removed.

Fixes #507 

## Proposed Changes

* Remove owner references on knative-local-gateway service after changing namespace to istio-system

### DIFF
--- a/pkg/reconciler/knativeserving/common/ingress_service.go
+++ b/pkg/reconciler/knativeserving/common/ingress_service.go
@@ -22,10 +22,13 @@ import (
 )
 
 // IngressServiceTransform pins the namespace to istio-system for the service named knative-local-gateway.
+// It also removes the OwnerReference to the operator, as they are in different namespaces, which is
+// invalid in Kubernetes 1.20+.
 func IngressServiceTransform() mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
 		if u.GetAPIVersion() == "v1" && u.GetKind() == "Service" && u.GetName() == "knative-local-gateway" {
 			u.SetNamespace("istio-system")
+			u.SetOwnerReferences(nil)
 		}
 		return nil
 	}


houshengbo:
@polivbr You made a good point. The `knative-local-gateway` Service also leads to some other issues, like https://github.com/knative/operator/issues/565.

pkg/reconciler/knativeserving/common/ingress_service.go
```diff
@@ -22,10 +22,13 @@ import (
 )
 
 // IngressServiceTransform pins the namespace to istio-system for the service named knative-local-gateway.
+// It also removes the OwnerReference to the operator, as they are in different namespaces, which is
+// invalid in Kubernetes 1.20+.
 func IngressServiceTransform() mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
 		if u.GetAPIVersion() == "v1" && u.GetKind() == "Service" && u.GetName() == "knative-local-gateway" {
 			u.SetNamespace("istio-system")
+			u.SetOwnerReferences(nil)
```
maximilien, line 31: BTW, I can comment this line 31 and all tests passing... so clearly need to add test to cover this chance. Thx.

maximilien (changes requested):
See comment.

houshengbo:
@polivbr @maximilien 
I am thinking of accepting this PR by now, coz we are close to the release of 0.23. Would like to get this issue resolved.
In terms of test cases, it is saved in my `cache memory`.

houshengbo:
https://github.com/knative/operator/issues/586 TO be resolved.
/lgtm
/approve

polivbr:
I'm stuck with getting my company's legal department to sign off on the CLA.

It's such a trivial fix... can someone else just replicate it if the lack of a corporate CLA is an issue?

maximilien:
> In terms of test cases, it is saved in my `cache memory`.

Or better, create an issue for it :)

PR #571, opened 2021-04-30
Allow for all Ingresses to be disabled

### LINKED ISSUE #568: No manifest is applied if all Ingresses are disabled
**Describe the bug**
If an Ingress config is defined, but all Ingress options are disabled, nothing gets installed at all.

**Expected behavior**
Knative gets installed but without any ingress.

**To Reproduce**
Apply the following YAML

```yaml
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  ingress:
    istio:
      enabled: false
    kourier:
      enabled: false
    contour:
      enabled: false
```

Note how nothing gets installed.

**Knative release version**

v0.21 and v0.22

**Additional context**

This is a bug in our defaulting and usage of `mf.Any` here: https://github.com/knative/operator/blob/39d7e473b24ae08101d440b43145d2d135a71fae/pkg/reconciler/knativeserving/ingress/ingress.go#L43-L59. The fix should be fairly straightforward.

/assign

markusthoemmes:
Fixes #568

## Proposed Changes

As per title (and as described in the issue) this allow for all ingresses to be disabled while still being able to deploy Knative Serving successfully.

**Release Note**

```release-note
Fixed a bug where disabling all Ingresses would break the installation of Knative Serving.
```

/assign @houshengbo

### DIFF
--- a/pkg/reconciler/knativeserving/ingress/ingress.go
+++ b/pkg/reconciler/knativeserving/ingress/ingress.go
@@ -40,11 +40,17 @@ func ingressFilter(name string) mf.Predicate {
 	}
 }
 
+// noneFilter drops all ingresses but allows everything else.
+func noneFilter(u *unstructured.Unstructured) bool {
+	_, hasLabel := u.GetLabels()[providerLabel]
+	return !hasLabel
+}
+
 // Filters makes sure the disabled ingress resources are removed from the manifest.
 func Filters(ks *v1alpha1.KnativeServing) mf.Predicate {
 	var filters []mf.Predicate
 	if ks.Spec.Ingress == nil {
-		return mf.Any(istioFilter)
+		return istioFilter
 	}
 	if ks.Spec.Ingress.Istio.Enabled {
 		filters = append(filters, istioFilter)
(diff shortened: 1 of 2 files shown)


maximilien (commented):
Nice. Quick question @markusthoemmes, what’s the use case for no ingress? Is it to deploy your own after? Or enable post install. Just curious. Thanks 🙏🏽

markusthoemmes:
@maximilien I could imagine both. The concrete use-case that made me run into this is that we're currently still installing Kourier outside of the operator downstream (because it predates the Ingress API), but thought that this behavior seemed kinda odd in a generalized setting.

pkg/reconciler/knativeserving/ingress/ingress.go
```diff
@@ -40,11 +40,17 @@ func ingressFilter(name string) mf.Predicate {
...
+	_, hasLabel := u.GetLabels()[providerLabel]
+	return !hasLabel
+}
+
 // Filters makes sure the disabled ingress resources are removed from the manifest.
 func Filters(ks *v1alpha1.KnativeServing) mf.Predicate {
 	var filters []mf.Predicate
 	if ks.Spec.Ingress == nil {
-		return mf.Any(istioFilter)
+		return istioFilter
```
houshengbo, line 53: What is the difference between mf.Any(istioFilter) and istioFilter? They both return the `istioFilter`.

markusthoemmes: They both return the **outcome** of `istioFilter`. This is just more lightweight as the Any wrapper is useless in this case.

houshengbo (commented):
/lgtm
/approve

The bug is valid and this fix resolves the issue.

PR #573, opened 2021-05-05
WIP: Adding two more components for scaling

matzew:

Fixes #

## Proposed Changes

* add the `filter` and `ingress` to the `haSupport` deployments

**Release Note**

```release-note

```

### DIFF
--- a/pkg/reconciler/common/ha.go
+++ b/pkg/reconciler/common/ha.go
@@ -29,7 +29,7 @@ const (
 	configMapName           = "config-leader-election"
 	enabledComponentsKey    = "enabledComponents"
 	servingComponentsValue  = "controller,hpaautoscaler,certcontroller,istiocontroller,nscontroller"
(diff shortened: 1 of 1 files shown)


aliok:
I don't see leader election anywhere in https://github.com/knative/eventing/blob/6136d09fb405a605c9239f3abaf2b1ec6e5cf5e2/cmd/broker/filter/main.go#L61

Are we sure HA is enabled for broker filter? Same story for broker ingress.

aliok:
What I mean by "" is, for example, mt broker controller is calling https://github.com/knative/eventing/blob/b482c5fad781520e3e3e0078159b4d786f408749/vendor/knative.dev/pkg/injection/sharedmain/main.go#L195 via `sharedmain.Main` here: https://github.com/knative/eventing/blob/6136d09fb405a605c9239f3abaf2b1ec6e5cf5e2/cmd/mtchannel_broker/main.go#L34

matzew:
/close
args - I thought we had something there ... ok. until than, I close

Issue #586, opened 2021-05-12
Add a test case to check the nil OwnerReferences on knative-local-gateway Service 


houshengbo:
To avoid cross-ns ownership, we need a test case to verify the knative-local-gateway Service is nil in owner reference.

PR #590, opened 2021-05-13
Set the namespace of the service knative-local-gateway to the istio namespace

### LINKED ISSUE #565: Knative operator looks for "istio-system" namespace when custom istio gateway is set

houshengbo:
Fixes #565 #586

## Proposed Changes

* Set the namespace of istio to the service knative-local-gateway based on the serving CR.

### DIFF
--- a/pkg/reconciler/knativeserving/common/ingress_service.go
+++ b/pkg/reconciler/knativeserving/common/ingress_service.go
@@ -25,11 +25,6 @@ import (
 	"knative.dev/operator/pkg/apis/operator/v1alpha1"
 )
 
-const (
-	IstioName       = "istio"
-	ConfigIstioName = "config-istio"
-)
-
 // IngressServiceTransform pins the namespace to istio-system for the service named knative-local-gateway.
 // It also removes the OwnerReference to the operator, as they are in different namespaces, which is
 // invalid in Kubernetes 1.20+.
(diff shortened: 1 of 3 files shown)


pkg/reconciler/knativeserving/common/ingress_service.go
```diff
@@ -17,19 +17,45 @@ limitations under the License.
...
 // It also removes the OwnerReference to the operator, as they are in different namespaces, which is
 // invalid in Kubernetes 1.20+.
-func IngressServiceTransform() mf.Transformer {
+func IngressServiceTransform(ks *v1alpha1.KnativeServing) mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
 		if u.GetAPIVersion() == "v1" && u.GetKind() == "Service" && u.GetName() == "knative-local-gateway" {
 			u.SetNamespace("istio-system")
 			u.SetOwnerReferences(nil)
+			config := ks.GetSpec().GetConfig()
+			if data, ok := config["istio"]; ok {
```
maximilien, line 37: What if the `config[“istio”]` is not present? Or could that be the case when different ingress is used?

maximilien (commented):
/ok-to-test

pkg/reconciler/knativeserving/common/ingress_service.go
houshengbo: if neither `istio` nor `config-istio` exists, we will use the default ns `u.SetNamespace("istio-system")`

maximilien (approved):
/approve
/lgtm

Issue #606, opened 2021-05-24
Unstable upgrading tests for eventing


houshengbo:
PLZ refer to the following links for the detailed error messages:

https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_operator/603/pull-knative-operator-eventing-upgrade-tests/1396876663534391296/build-log.txt

https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_operator/603/pull-knative-operator-eventing-upgrade-tests/1396876663534391296

It is likely to fail the test case as shown in the above links, when upgrading eventing from 0.23 to the latest.

Upgrading from 0.22 to 0.23, or across any other prior versions, the test case does not fail as frequently as from 0.23 to latest.

houshengbo:
In the upgrading test cases for eventing CR, it verifies if the eventing CR is READY and all the deployment resources are AVAILABLE with the correct target version. After they are all up and running, it launches the the post upgrade tests in eventing repository. However, the above error happens almost 50%-60%. The last time I retried with the prow, I got 3 times failure out of 6.

Then, I added a timeout of 3 minutes after upgrading with the eventing CR, and I got none prow failing out of 6, during the last time I retried with the prow.

It is fair for me to believe that eventing now needs a `magic timestamp point` for things to settle down for upgrading, but still have no idea what the indicator is for the magic timestamp point. What can indicate a complete finish of the eventing upgrade?

houshengbo:
@cardil @vaikas @lionelvillard

cardil:
The failures are the simplest smoke tests that are being run after eventing upgrade. It's boggling why they fail.

Latest changes of eventing upgrade tests didn't touch them, as I was focused on expendability of continual tests (running in background).

I must say that during my latest work on eventing upgrades, I think I did observe similar behavior of those smoke tests at some point, but it went away after some merge. Now, it behaves stable.

There is observed instability of continual tests, but it's related to lost events on IMC running with retries: https://github.com/knative/eventing/issues/5430. Post upgrade tests are running stable.

houshengbo:
@cardil It is possible that this failure went away, since eventing upgrade has a higher chance to beat the post upgrade tests in eventing repo, because eventing repo applies the yamls directly, which takes less time than the process in operator.

[...]

PR #623, opened 2021-05-27
Add a 20-second timeout before running the post upgrade tests

houshengbo:
Fixes: #606 

Also refactors some of the wait-for-ready logic in the tests in hopes to provide clearer error signals.

Picked up a new version of pkg.

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -263,6 +263,7 @@ metadata:
 spec:
   version: "${TARGET_RELEASE_VERSION}"
 EOF
+  sleep 5m
 }
 
 function if_version_exists() {
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -263,11 +263,6 @@ metadata:
 spec:
   version: "${TARGET_RELEASE_VERSION}"
 EOF
-  # Operator has the issue making sure all deployments are ready before running the postupgrade
-	# tests, especially when spec.version is set to latest. Before figuring out the optimal approach,
-	# we add a timeout of 20 seconds here to make sure all the deployments are up and running for the
-	# target version.
-	sleep 20s
 }
 
 function if_version_exists() {
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -263,7 +263,6 @@ metadata:
 spec:
   version: "${TARGET_RELEASE_VERSION}"
 EOF
-  sleep 5m
 }
 
 function if_version_exists() {
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -263,6 +263,11 @@ metadata:
 spec:
   version: "${TARGET_RELEASE_VERSION}"
 EOF
+  # Operator has the issue making sure all deployments are ready before running the postupgrade
+	# tests, especially when spec.version is set to latest. Before figuring out the optimal approach,
+	# we add a timeout of 20 seconds here to make sure all the deployments are up and running for the
+	# target version.
+	sleep 20s
 }
 
 function if_version_exists() {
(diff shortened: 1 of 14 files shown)


test/resources/knativeresources.go
```diff
@@ -91,7 +91,24 @@ func IsKnativeDeploymentReady(dpList *v1.DeploymentList, expectedDeployments []s
...
+
+			if key == "networking.knative.dev/ingress-provider" {
+				statusCheck = true
+			}
+
+			if version == common.LATEST_VERSION && version == existingVersion {
+				statusCheck = true
+			}
+
+			if version == common.LATEST_VERSION && version != existingVersion && (key == "serving.knative.dev/release" || key == "eventing.knative.dev/release") && val != fmt.Sprintf("v%s", existingVersion) {
```
evankanderson, line 107: I don't quite understand the extra logic here -- might be worth a comment, or rewriting as:

```golang
if version == common.LATEST_VERSION && version != existingVersion {
  if (key == "serving.knative.dev/release" || key = "eventing.knative.dev/release") && val != fmt.Sprintf("v%s", existingVersion) {
    statusCheck = true
  }
}
```

There's such a variety of cases here, I'm struggling to figure out what the common axes are. Reading a bit further, it looks like we know a `version` and `existingVersion` from some prior knowledge, and are checking labels for matches. It feels like we should have an early-exit on the loop for labels where the key doesn't match, e.g.

```golang
for k, v := range d.GetObjectMeta().GetLabels() {
  if key == "networking.knative.dev/ingress-provider" {
    break
  }

  if key == "serving.knative.dev/release" || key == "serving.knative.dev/release" {
    // If versions don't match
    return false
  }
}
return isReady(d.Status)  // At least check the Deployment conditions, but there might be other "ready" statuses that need checking.

See https://carvel.dev/kapp/docs/latest/apply-waiting/#overview for some other examples of "waiting" things (`batch/v1/Job` in particular).

Comment (evankanderson) [reply test/resources/knativeresources.go:107]: In particular, I'm concerned about:

```golang
if val == fmt.Sprintf("v%s", version) && version != common.LATEST_VERSION
```

possibly being able to pick up unexpected label keys.

evankanderson (commented):
I could fairly easily approve the "add 3 minutes" for now to get Prow unstuck, but there's a bunch of other logic threading through "existingVersion" that I needed to look at; `IsKnativeDeploymentReady` feels like a layering violation as well as a bit of complexity that might want to be factored out. I.e. maybe we should check the `Ready` status of all the appropriate CRs, rather than the Deployment (which might still need to read configuration even after the Pods have finished rollout).

houshengbo:
@evankanderson @maximilien I shorten the timeout to 20 seconds. The prow works fine so far so good. No failure is found.

test/e2e-common.sh
```diff
@@ -263,6 +263,11 @@ metadata:
 spec:
   version: "${TARGET_RELEASE_VERSION}"
 EOF
+  # Operator has the issue making sure all deployments are ready before running the postupgrade
+	# tests, especially when spec.version is set to latest. Before figuring out the optimal approach,
+	# we add a timeout of 20 seconds here to make sure all the deployments are up and running for the
+	# target version.
+	sleep 20s
```
cardil, line 270: Instead of adding it here, I recommend adding additional `PostUpgradeOperation` that will emit appropriate warn and wait for this amount.

test/resources/knativeresources.go
houshengbo: I will add comments as many as possible, but I would like to keep the existing logic.

test/e2e-common.sh
houshengbo: done

evankanderson:
I've updated the PR to explain the 12 other files added.

/lgtm
/approve

Issue #5, opened 2020-03-31
Operations: Please allow setting nodeSelector, tolerations, and affinity on all Pods


AceHack:
Most companies with complex enough kubernetes deployments that have multiple different node groups require some sort of topology selection so pods will run on the correct group of nodes.  Please allow setting nodeSelector, tolerations, and affinity on all Pods to accomplish this.  This includes adding it to any CRDs that create pods so these values are settable.

https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/

See below for example:
```
apiVersion: v1
kind: Pod
metadata:
  name: sample
spec:
  nodeSelector:
    disktype: sad
  tolerations:
  - key: "example-key"
    operator: "Exists"
    effect: "NoSchedule"
  affinity:
    nodeAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 1
        preference:
          matchExpressions:
          - key: another-node-label-key
            operator: In
            values:
            - another-node-label-value
```

Related Feature Request: https://github.com/knative/eventing/issues/2883

houshengbo:
This is what is implemented in other operator:

NodeSelector: https://github.com/helm/charts/blob/master/stable/prometheus-operator/crds/crd-prometheus.yaml#L3223

Affinity: https://github.com/helm/charts/blob/master/stable/prometheus-operator/crds/crd-prometheus.yaml#L119

Tolerations: https://github.com/helm/charts/blob/master/stable/prometheus-operator/crds/crd-prometheus.yaml#L4523

AceHack:
We found PodNodeSelector and PodTolerationRestriction admission controllers.
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podtolerationrestriction
https://github.com/kubernetes/kubernetes/issues/57424

But most hosted K8s control plane providers do not support them because their annotations are alpha.
https://docs.aws.amazon.com/eks/latest/userguide/platform-versions.html

We also found this open-source project that allows for setting node selector in namespace annotations using a webhook.
https://github.com/liangrog/admission-webhook-server

We will likely try to contribute the other tolerations, and affinity back to this project assuming the maintainer is open to it.

AceHack:
[...]

PR #658, opened 2021-06-10
Allow to set NodeSelector via `spec.deployments.nodeSelector`

nak3:

This patch adds `spec.deployments.nodeSelector` to set nodeSelector on
each deployment.

For example, when the following CR is created,
```
kind: KnativeServing
metadata:
  name: ks
  namespace: knative-serving
spec:
  high-availability:
    replicas: 1
  deployments:
  - name: webhook
    nodeSelector:
      foo: ba
```

The webhook deployment has `spec.template.spec.nodeSelector`.

```
$ kubectl get deploy -n knative-serving  webhook -o jsonpath={.spec.template.spec.nodeSelector}
{"foo":"bar"}
```

/cc @houshengbo @markusthoemmes @matzew

### DIFF
--- a/pkg/reconciler/common/deployments_override.go
+++ b/pkg/reconciler/common/deployments_override.go
@@ -43,6 +43,7 @@ func DeploymentsTransform(obj v1alpha1.KComponent, log *zap.SugaredLogger) mf.Tr
 				replaceLabels(&override, deployment)
 				replaceAnnotations(&override, deployment)
 				replaceReplicas(&override, deployment)
+				replaceNodeSelector(&override, deployment)
 				if err := scheme.Scheme.Convert(deployment, u, nil); err != nil {
 					return err
 				}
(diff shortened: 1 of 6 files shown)


markusthoemmes (approved):
Looks spot on to me!

/hold

For feedback from others.

pkg/reconciler/common/deployments_override.go
```diff
@@ -86,3 +87,9 @@ func replaceReplicas(override *v1alpha1.DeploymentOverride, deployment *appsv1.D
 		deployment.Spec.Replicas = &override.Replicas
 	}
 }
+
+func replaceNodeSelector(override *v1alpha1.DeploymentOverride, deployment *appsv1.Deployment) {
+	if len(override.NodeSelector) > 0 {
+		deployment.Spec.Template.Spec.NodeSelector = override.NodeSelector
```
houshengbo, line 93: This feature is fine.
My question is whether to overwrite and add the nodeselector.
NodeSelector is a key-value map, the same to Labels and Annotations.
For Labels and Annotations, we add the override into the existing map, but for NodeSelector, we replace the existing with the override. Why not keep the existing keys, that is not in override for NodeSelector?

markusthoemmes: I don't think it's useful to merge node selectors. IIRC they're in an `AND` combination, so adding more would likely result in no nodes matching anymore. Plus: We don't have preset node selectors at all.

nak3: Yes I made this by design because of the reason Markus mentioned.

houshengbo: Then I am ok with this PR.

nak3:
I will update the doc here https://knative.dev/docs/install/operator/configuring-serving-cr/#override-system-deployments

PR #661, opened 2021-06-11
Allocate the array with a smaller initial size

houshengbo:


pkg/reconciler/knativeeventing/source/source.go
```diff
@@ -54,7 +54,7 @@ func getSourcePath(version string, ke *v1alpha1.KnativeEventing) string {
 
 	// This line can make sure a valid available source version is returned.
 	sourcePath := filepath.Join(koDataDir, "eventing-source", sourceVersion)
-	urls := make([]string, 0, 100)
+	urls := make([]string, 0)
```
maximilien, line 57: I always wonder if `urls := []string{}` are better style. But either way.

maximilien (approved):
/ok-to-test
/approve
/lgtm

pkg/reconciler/knativeeventing/source/source.go
markusthoemmes: In this case, it should actually be `var urls []string` to avoid an allocation if there's no source enabled.

/hold

julz: Not that it matters literally at all, but I’d do `var urls []string` here, the difference is incredibly small, but the var version doesn’t needlessly create the array if it doesn’t end up getting used.

julz: heh oops, @markusthoemmes said the same thing :)

markusthoemmes (approved):
/lgtm
/approve
/unhold

PR #674, opened 2021-06-28
Install the webhooks after installing the deployments and services

houshengbo:
## Proposed Changes

* The webhooks rely on the services. The webhooks can be triggered during the installation of the manifests, because CMs are installed as well. We need to make sure that the services webhooks rely on, are installed before the installation of the webhooks.
* It may lead to some issues during the upgrading, if the webhooks are installed before the services. There is a gap the webhook is updated, but the service is not available. During this time, a change on the CM can trigger the webhook, raising errors.
* Update the manifests and dependencies for the main branch.


maximilien (approved):
/ok-to-test
/approve
/lgtm

maximilien (approved):
/lgtm

markusthoemmes:
Why are there so many manifest changes in this PR? Seems to do much more than the title implies.

PR #734, opened 2021-08-28
Use `knative.dev/hack/codegen-library.sh` in `hack/update-codegen.sh`

nak3:
operator still does not use the common script knative.dev/hack/codegen-library.sh
so update deps failed as https://github.com/knative/operator/pull/726.

This patch fixes the script and includes https://github.com/knative/operator/pull/726.

/cc @houshengbo @evankanderson @mattmoor


evankanderson (commented):
/lgtm
/approve

PR #735, opened 2021-08-28
Drop deprecated `enabledComponents` field setting

nak3:
This patch removes deprecated `enabledComponents` field in config-leader-election

It was removed since 1 year ago https://github.com/knative/pkg/pull/2162.

/cc @houshengbo @evankanderson @maximilien


nak3:
The deps issue shoould be fixed by https://github.com/knative/operator/pull/734.

PR #740, opened 2021-08-31
Remove pingsource from HA scaling

matzew:

Fixes #

## Proposed Changes

* the pingsource is scaled, based on demand, via Knative Eventing itself - Instead of hard-wired via HA config. Therefore removing form the HA apply section

**Release Note**

```release-note
PingSource no langer is having a replica as default. It is scaled on demand
```

### DIFF
--- a/pkg/reconciler/common/ha.go
+++ b/pkg/reconciler/common/ha.go
@@ -44,7 +44,6 @@ func haSupport(obj v1alpha1.KComponent) sets.String {
 		"imc-controller",
 		"imc-dispatcher",
 		"mt-broker-controller",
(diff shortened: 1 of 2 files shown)


houshengbo (commented):
/lgtm
/approve

lionelvillard:
AFAIK, PingSource is not scaled on-demand. Is this new?

matzew:
@lionelvillard the adapter is created on demand, when we actually create a `PingSource`, it scaled up: 
https://github.com/knative/eventing/blob/main/config/core/deployments/pingsource-mt-adapter.yaml#L24

that's what I meant

PR #743, opened 2021-08-31
Add the lables: app.kubernetes.io/version and app.kubernetes.io/part-of

### LINKED ISSUE #681: Adopt `app.kubernetes.io` labels
The operator needs to apply the following labels to all k8s resources it creates as well as pod selectors:

```
app.kubernetes.io/name
app.kubernetes.io/version
app.kubernetes.io/component
```

Part of https://github.com/knative/community/issues/667

app and version labels will be deprecated at a later date.

houshengbo:
Fixes #681

### DIFF
--- a/config/operator.yaml
+++ b/config/operator.yaml
@@ -19,6 +19,7 @@ metadata:
   namespace: default
   labels:
     operator.knative.dev/release: devel
+    app.kubernetes.io/name: knative-operator
     app.kubernetes.io/version: devel
     app.kubernetes.io/part-of: knative-operator
 spec:
@@ -32,6 +33,9 @@ spec:
         sidecar.istio.io/inject: "false"
       labels:
         name: knative-operator
+        app.kubernetes.io/name: knative-operator
+        app.kubernetes.io/version: devel
+        app.kubernetes.io/part-of: knative-operator
     spec:
       serviceAccountName: knative-operator
       containers:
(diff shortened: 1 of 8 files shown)


config/operator.yaml
```diff
@@ -19,6 +19,8 @@ metadata:
   namespace: default
   labels:
     operator.knative.dev/release: devel
+    app.kubernetes.io/version: devel
```
upodroid, line 22: Can you add `app.kubernetes.io/name: knative-operator` as well? The `name` key is added to Deployments/STSs/DSs and Services.

Also, add the new labels to `spec.template.metadata.labels`.

Can you append this Service too?

```
---
apiVersion: v1
kind: Service
metadata:
  labels:
    serving.knative.dev/release: devel
    app.kubernetes.io/name: knative-operator
    app.kubernetes.io/version: devel
    app.kubernetes.io/part-of: knative-operator
  name: knative-operator
  namespace: knative-operator
spec:
  ports:
  # Define metrics for it to be accessible within service meshes.
  - name: http-metrics
    port: 9090
    targetPort: 9090
  selector:
    name: knative-operator
```

Also, why is Istio injection disabled on this deployment?

upodroid (changes requested):

config/operator.yaml
houshengbo: I do not know why istio is disabled. it has been there since the beginning.
Why do we need to add this service?

upodroid: It is to create a servicemonitor targeting the operator pods

houshengbo: Let's open another issue about the connection to monitoring.

upodroid: Ok

PR #745, opened 2021-08-31
fix: add status error while version is not supported

### LINKED ISSUE #741: No error message in `status` if the `spec.version` is not supported
**Describe the bug**
Operator version: v0.25.0
KnativeServing's `spec.version`: v0.21.0
There is error logs in operator controller, but error is not shown in CR's status.

**Expected behavior**
Show error log in CR's status.

**To Reproduce**
1. install knative-operator with version v0.25.0
2. apply KnativeServing with spec.version=0.21.0 and spec.manifests must not be nil
3. kubectl logs controller-pod-xxx...

**Knative release version**
v0.25.0

**Additional context**

charles-woshicai:
Signed-off-by: charles-woshicai <583605554@qq.com>

Fixes #741 

## Proposed Changes

*
*
*

**Release Note**

```release-note

```

### DIFF
--- a/pkg/reconciler/common/stages.go
+++ b/pkg/reconciler/common/stages.go
@@ -18,6 +18,7 @@ package common
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	mf "github.com/manifestival/manifestival"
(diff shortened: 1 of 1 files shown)


charles-woshicai:
@googlebot I signed it!

pkg/reconciler/common/stages.go
```diff
@@ -110,6 +111,11 @@ func DeleteObsoleteResources(ctx context.Context, instance v1alpha1.KComponent,
 	installed, err := fetch(ctx, instance)
 	if err != nil {
 		logger.Error("Unable to obtain the installed manifest; obsolete resources may linger", err)
+		// Report an error in `Status`, issue ref: https://github.com/knative/operator/issues/741
```
houshengbo, line 114: I do not think this is the best place to add the message in the status.
Since we set spec.version which is not supported by the current version, we need to raise the error, where we check the spec.version and get the target manifests.
And report "the target version is too old, unable to get the target manifests or something".

charles-woshicai: Yeah I agree with you, that's exactly why I ask whether we should reject in the issue.

maximilien (commented):
Running go format on your code should remove these warnings…

pkg/reconciler/common/stages.go
houshengbo: https://github.com/knative/operator/pull/750  I opened another PR. I think that's a better solution to resolve this issue.
I users specify spec.manifests with the version, operator will allow that.

charles-woshicai:
Close it for there is solution #750

PR #747, opened 2021-09-01
Allow to set Tolerations via `spec.deployments.tolerations`

nak3:

This patch adds `spec.deployments.tolerations` to set tolerations on
each deployment.

For example, when the following CR is created,
```
kind: KnativeServing
metadata:
  name: ks
  namespace: knative-serving
spec:
  high-availability:
    replicas: 1

  deployments:
  - name: webhook
    tolerations:
    - effect: NoSchedule
      key: node-type
      operator: Equal
      value: production
```

The webhook deployment has `spec.template.spec.tolerations`.

```
$ kubectl get deploy webhook -o jsonpath={.spec.template.spec.tolerations}
[{"effect":"NoSchedule","key":"node-type","operator":"Equal","value":"production"}]
```

/cc @houshengbo

### DIFF
--- a/config/300-eventing.yaml
+++ b/config/300-eventing.yaml
@@ -104,6 +104,46 @@ spec:
                         type: string
                       description: NodeSelector overrides nodeSelector for the deployment.
                       type: object
+                    tolerations:
+                      description: If specified, the pod's tolerations.
+                      items:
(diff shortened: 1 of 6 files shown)


houshengbo (commented):
/lgtm
/approve

PR #749, opened 2021-09-01
Apply `high-availability` to  all deployments managed by operator

nak3:
Fix https://github.com/knative/operator/issues/337

Currently `high-availability` does not apply to all Deployments but
deployments listed in `haSupport()` only.

All components in Knative project should support HA (please correct me
if I was wrong) and so `high-availability` should also support
all deployment's HA.

__if__  there are some components which do not support HA, we should
create `haUnSupport()` fund and add it there.

**Release Note**

```release-note
`high-availability` applies all deployments managed by operator
```

/cc @houshengbo

### DIFF
--- a/pkg/reconciler/common/ha.go
+++ b/pkg/reconciler/common/ha.go
@@ -24,7 +24,7 @@ import (
 	v1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
 )
 
-func unSupported(obj v1alpha1.KComponent) sets.String {
+func haUnSupported(obj v1alpha1.KComponent) sets.String {
 	return sets.NewString(
 		"pingsource-mt-adapter",
 	)
@@ -50,7 +50,7 @@ func HighAvailabilityTransform(obj v1alpha1.KComponent, log *zap.SugaredLogger) 
 		replicas := int64(ha.Replicas)
 
 		// Transform deployments that support HA.
-		if u.GetKind() == "Deployment" && !unSupported(obj).Has(u.GetName()) {
+		if u.GetKind() == "Deployment" && !haUnSupported(obj).Has(u.GetName()) {
 			if err := unstructured.SetNestedField(u.Object, replicas, "spec", "replicas"); err != nil {
 				return err
 			}
--- a/pkg/reconciler/common/ha.go
+++ b/pkg/reconciler/common/ha.go
@@ -20,32 +20,8 @@ import (
 	mf "github.com/manifestival/manifestival"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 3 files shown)


nak3:
/hold

I believe that all eventing's component support HA but it is better to confirm it on Slack or somewhere.

houshengbo:
@evankanderson @dprotaso @lionelvillard @maximilien @matzew 
What do you talents think?

pkg/reconciler/common/ha.go
```diff
@@ -24,26 +24,9 @@ import (
...
-		"net-certmanager-controller",
-		"net-istio-controller",
-		"net-kourier-controller",
-		"net-kourier-gateway",
-		"eventing-controller",
-		"sugar-controller",
-		"imc-controller",
-		"imc-dispatcher",
-		"mt-broker-controller",
+		"pingsource-mt-adapter",
```
matzew, line 29: this would than also make the webhook to be HA, right?

those are currently only HPA:
* https://github.com/knative/operator/blob/main/cmd/operator/kodata/knative-serving/0.25.1/2-serving-core.yaml#L5434-L5447
* https://github.com/knative/operator/blob/main/cmd/operator/kodata/knative-eventing/0.25.1/2-eventing-core.yaml#L846-L864

nak3: Yes, that's right.
webhook which has HPA is scaled up by the operator but other webhook are not be HA now. But all deployment including webhook will effect on the ha config.

matzew: so we need to exclude the webhook than? Or we want it included, as it was missing before?

nak3: I personally want to include the webhook. It is difficult to understand the current HA components that are effected by HA setting. So including all deployments and excluding unsupported components explicitly like `pingsource-mt-adapter` make it clear.

pkg/reconciler/common/ha.go
```diff
@@ -24,26 +24,9 @@ import (
 	v1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
 )
 
-func haSupport(obj v1alpha1.KComponent) sets.String {
+func unSupported(obj v1alpha1.KComponent) sets.String {
```
maximilien, line 27: Also is `unSupported` best name for this?

maximilien (commented):
/ok-to-test

Any way to test this?

nak3:
Yes, added e2e for HA. (It seems no e2e for KnativeEventing CR's config so added KnativeServing only for now.)

pkg/reconciler/common/ha.go
houshengbo: I like this approach. We change it from a W-list to a B-list.

lionelvillard: `pingsource-mt-adapter` supports HA.

houshengbo:
@nak3 Since I am preparing the new release, but need this PR badly.
I just merge it, and revise it later with the list of unsupported deployments. PlZ understand.
Thank you.

PR #750, opened 2021-09-02
Fix the error issue with installing spec.manifests

### LINKED ISSUE #741: No error message in `status` if the `spec.version` is not supported

houshengbo:
Fixes #741

If the spec.manifests is specified with an older version, not supported by the operator, the operator can still work by merely installing what is available in spec.manifests.

### DIFF
--- a/pkg/reconciler/knativeeventing/source/source.go
+++ b/pkg/reconciler/knativeeventing/source/source.go
@@ -104,7 +104,7 @@ func AppendTargetSources(ctx context.Context, manifest *mf.Manifest, instance v1
 	}
 	if len(instance.GetSpec().GetManifests()) != 0 {
 		// If spec.manifests is not empty, it is possible that the eventing source is not available with the
-		// specified version. The user can specify the ingress link in the spec.manifests.
+		// specified version. The user can specify the eventing source link in the spec.manifests.
 		return nil
 	}
 	return err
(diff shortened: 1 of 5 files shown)


houshengbo:
@charles-woshicai 
I tried to fix the issue like this.
spec.manifests with an older version should work.

pkg/reconciler/knativeeventing/source/source.go
```diff
@@ -103,7 +98,16 @@ func getSourcePath(version string, ke *v1alpha1.KnativeEventing) string {
...
 	version := common.TargetVersion(instance)
 	sourcePath := getSourcePath(version, convertToKE(instance))
-	return getSource(manifest, sourcePath)
+	m, err := getSource(manifest, sourcePath)
+	if err == nil {
+		*manifest = manifest.Append(m)
+	}
+	if len(instance.GetSpec().GetManifests()) != 0 {
+		// If spec.manifests is not empty, it is possible that the eventing source is not available with the
+		// specified version. The user can specify the ingress link in the spec.manifests.
```
lionelvillard, line 107: is this ingress link or source link?

PR #777, opened 2021-09-15
Add the support of `spec.deployments.affinity`

houshengbo:

This patch adds `spec.deployments.affinity` to set affinity on each deployment.


nak3:
/lgtm

I think this PR and the deps update should be separated but it seems https://github.com/knative/operator/pull/777/commits/07a21183275b23f215c280ba2b3c0ba517651735 (update CRD) included both :disappointed:

PR #863, opened 2021-11-10
Refactor image transformer to deduplicate and simplify it

markusthoemmes:
## Proposed Changes

The code here contained so many indirections, that it was IMO impossible to follow. This fixes that, gets rid of unnecessary interfaces etc. and straight up makes things easier to read.

/assign @houshengbo


aliok (approved):
/lgtm
/approve

PR #893, opened 2021-12-07
Add the support of resources configuration based on deploy and container names

### LINKED ISSUE #860: "controller" resources configuration overlap between "controller" and "net-istio-controller"
**Describe the bug**

https://knative.dev/docs/install/operator/configuring-serving-cr/#system-resource-settings
indicates

```Requests and limits can be configured for the following containers: activator, autoscaler, controller, webhook, autoscaler-hpa, net-istio-controller and queue-proxy.```

When in fact, net-istio-controller does not work (as in applying resources changes).   It has a container called 'controller' and not 'net-istio-controller'.

If you add a resources section for `controller` as in

```  
- container: controller
    limits:
      cpu: 300m
      memory: 1000Mi
    requests:
      cpu: 30m
      memory: 800Mi
```

it will apply the new resource to the `controller` deployment in addition to `net-istio-controller` which is undesirable. 

**Expected behavior**

Behavior as documented that would allow `net-istio-controller` and `controller` to have resources adjusted independently. 

**To Reproduce**

Try to change resources for `net-istio-controller` as described in the CR document above.

**Knative release version**
0.26

**Additional context**
N/A

### PR DESCRIPTION
**Pull request by houshengbo**
Fixes #860

### DIFF
--- a/config/300-serving.yaml
+++ b/config/300-serving.yaml
@@ -739,6 +739,47 @@ spec:
                               type: array
                           type: object
                       type: object
+                    resources:
+                      description: If specified, the container's resources.
+                      items:
+                        description: The pod this Resource is used to specify the requests and limits for
+                          a certain container based on the name.
+                        properties:
+                          container:
+                            description: The name of the container
+                            type: string
+                          limits:
+                            properties:
+                              cpu:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              ephemeral-storage:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              memory:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              storage:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                            type: object
+                          requests:
+                            properties:
+                              cpu:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              ephemeral-storage:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              memory:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              storage:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                            type: object
+                        type: object
+                      type: array
               ingress:
                 description: The ingress configuration for Knative Serving
                 properties:
--- a/config/300-serving.yaml
+++ b/config/300-serving.yaml
@@ -753,13 +753,7 @@ spec:
                               cpu:
                                 pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
                                 type: string
-                              ephemeral-storage:
-                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
-                                type: string
                               memory:
-                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
-                                type: string
-                              storage:
                                 pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
                                 type: string
                             type: object
@@ -768,13 +762,7 @@ spec:
                               cpu:
                                 pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
                                 type: string
-                              ephemeral-storage:
-                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
-                                type: string
                               memory:
-                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
-                                type: string
-                              storage:
                                 pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
                                 type: string
                             type: object
--- a/config/300-eventing.yaml
+++ b/config/300-eventing.yaml
@@ -718,6 +718,47 @@ spec:
                               type: array
                           type: object
                       type: object
+                    resources:
+                      description: If specified, the container's resources.
+                      items:
+                        description: The pod this Resource is used to specify the requests and limits for
+                          a certain container based on the name.
+                        properties:
+                          container:
+                            description: The name of the container
+                            type: string
+                          limits:
+                            properties:
+                              cpu:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: config/300-serving.yaml [hunk 1]
```diff
@@ -739,6 +739,47 @@ spec:
...
+                        properties:
+                          container:
+                            description: The name of the container
+                            type: string
+                          limits:
+                            properties:
+                              cpu:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              ephemeral-storage:
```
nak3, line 756: serving system pods do not use `ephemeral-storage` so this should be dropped.

config/300-serving.yaml
```diff
@@ -739,6 +739,47 @@ spec:
...
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              memory:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              storage:
```
nak3, line 762: Same with `ephemeral-storage`. It is not used.

config/300-serving.yaml
```diff
@@ -739,6 +739,47 @@ spec:
...
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                            type: object
+                          requests:
+                            properties:
+                              cpu:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              ephemeral-storage:
```
nak3, line 771: Same with above.

config/300-serving.yaml
```diff
@@ -739,6 +739,47 @@ spec:
...
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              memory:
+                                pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
+                                type: string
+                              storage:
```
nak3, line 777: Same with above.

nak3 (commented):
Will you not make `spec.resources.container` deprecated?
I think it is difficult to say that https://github.com/knative/operator/issues/860 is by design not a bug.

houshengbo:
@nak3 `spec.resources.container` will still work as before.
This pr just gives another way to configure certain containers under a specific deployment.

config/300-serving.yaml
houshengbo: it is defined in spec.resources already as well.

config/300-serving.yaml
houshengbo: it is defined in spec.resources already as well.

nak3:
Hmm... so, what situation users should use `spec.resources.container` instead of `spec.deployments.resources`?
`net-istio-controller` and serving's `controller` are **different controllers** but they just started having the same container name since **recent release**.
Even though `spec.deployments.resources` can cover `spec.resources.container` functionality, why we don't want to deprecate `spec.resources.container`?

houshengbo:
@nak3 It is possible that the user just want to configure all the containers named "controller" with the same resource limits.
In addition, what do you mean by deprecate? Do you mean we will gradually remove it? We have just released 1.0, and should be discreet with the field deletion in the CRD.

matzew:
why does 1.0.0 imply that we can not change APIs, moving forward?

rhuss:
> It is possible that the user just want to configure all the containers named "controller" with the same resource limits.

And what if not? If she wants to change same-named containers individually?

config/300-serving.yaml
rhuss: but they were never evaluated or used, so starting to ignore them would not harm as they have been already ignored all the time (if already given). So removing it is not really removing a field, it's a clean-up. If someone has used that, then for some reason and she would be surprised that it has no effect at all.

I'm +1 for only adding the fields that are actually used, and adding more in the future if needed.

houshengbo:
@rhuss That's exactly what this PR implements. If they want to configure individually, do it with the change in this pr by specifying the deploy and container name.

houshengbo:
@matzew I just do not want to bring the changes to drastically at this moment, like deleting fields or renaming the fields. Both of them require operator has a mature mechanism to convert forwards and backwards to remain compatibility.
As I said, operator needs to implement the webhook first to do that, which we do not have by now.

nak3:
> @nak3 It is possible that the user just want to configure all the containers named "controller" with the same resource limits.

Hmm... I have to say that it is strange design rather than good design... Again, each controller is different controller but just having the same container name.

> In addition, what do you mean by deprecate? Do you mean we will gradually remove it?

You can refer to the deprecation process in the docs.
https://github.com/knative/community/blob/main/mechanics/RELEASE-VERSIONING-PRINCIPLES.md#overarching-principle
https://kubernetes.io/docs/reference/using-api/deprecation-policy/

Anyway, if you think webhook can solve the problem, you should implement the webhook before this PR.
If you support both `spec.deployments.resources` and `spec.resources.container` fields, users must need extra care - understand what are different, do not set both fields, etc.

houshengbo:
Both spec.deployments.resources  and spec.container.resources can be used together, but spec.deployments.resources  will take precedence, if they configure the same fields.

nak3:
> Both spec.deployments.resources and spec.container.resources can be used together, but spec.deployments.resources will take precedence, if they configure the same fields.

I don't think it is true. If you want to do that, you need a code like [this](https://github.com/knative/operator/blob/3d2e015b311d6e12db2538f7297cdab3b1433c52/pkg/reconciler/common/ha.go#L38-L42).
 
> This can be a quick fix for 1.1. We can deal with the schema change later.

As you are thinking you need to discreet with the field deletion in the CRD, it is better to deal with the schema change before adding a new field, isn't it?

houshengbo:
>> Both spec.deployments.resources and spec.container.resources can be used together, but spec.deployments.resources will take precedence, if they configure the same fields.

> If you want to do that, you need a code like [this](https://github.com/knative/operator/blob/3d2e015b311d6e12db2538f7297cdab3b1433c52/pkg/reconciler/common/ha.go#L38-L42). Otherwise, operator is churned.

I will add that for sure. 
 
>> This can be a quick fix for 1.1. We can deal with the schema change later.

> As you are thinking you need to discreet with the field deletion in the CRD, it is better to deal with the schema change before adding a new field, isn't it?

Adding a field is different from deleting or renaming a field. When the new CRD schema overwrites the existing old one during the upgrade, the existing CR will be converted into the new CR without any issue. When downgrading to an older version of knative component, operator still uses the new CRD, as there are three version prior to the latest version supported by the operator.

For new CRD shema deleting or renaming a field, the upgrading of the existing CR could break, after upgrading the operator to the latest version. However, this would not happen for adding a field.

nak3:
Nah, I mean, "We still cannot deal with how to delete but let's add a new field" is not good idea... The https://github.com/knative/operator/issues/860 is not a new bug and a critical issue. I don't understand why you are so rush to push the change in v1.1...

And I'm sorry but it is difficult for me to agree some of the design and change process... Could you please ask someone else to review this? I'm sorry.

houshengbo:
I am not rushing anything.
I just hope the operator can improve, as we all do.
If it can be be settled down now, that's perfectly fine. I did not even know THIS WAS an ISSUE, until the recent discussion in the slack. I thought configuration on the container name level was good enough already.

maximilien (approved):
/approve
/lgtm

rhuss:
> This can be a quick fix for 1.1. We can deal with the schema change later.

@houshengbo please link the corresponding tracking issue to this PR.

houshengbo:
@rhuss Fixes #860 It is linked in the comment.

rhuss:
@houshengbo I mean what do you mean with "We can deal with the schema change later". Is there a tracking issue for this planned future addition?

Issue #970, opened 2022-02-19
Integration tests failed due to the conversion webhook


houshengbo:
```
=== RUN   TestKnativeServingDeployment
    knativeservingdeployment_test.go:44: KnativeService "knative" failed to create: conversion webhook for operator.knative.dev/v1alpha1, Kind=KnativeServing failed: Post "[https://operator-webhook.default.svc:443/resource-conversion?timeout=30s](https://operator-webhook.default.svc/resource-conversion?timeout=30s)": EOF
--- FAIL: TestKnativeServingDeployment (0.21s)
FAIL test/e2e.TestKnativeServingDeployment (0.21s)
FAIL test/e2e
=== Failed
=== FAIL: test/e2e TestKnativeServingDeployment (0.21s)
    knativeservingdeployment_test.go:44: KnativeService "knative" failed to create: conversion webhook for operator.knative.dev/v1alpha1, Kind=KnativeServing failed: Post "[https://operator-webhook.default.svc:443/resource-conversion?timeout=30s](https://operator-webhook.default.svc/resource-conversion?timeout=30s)": EOF
DONE 5 tests, 1 failure in 160.631s
```

PR #974, opened 2022-02-22
Fix the issue with alpha and beta transition

houshengbo:
Fixes: #970
KnativeServing transition between v1alpha1 and v1beta1 used to fail due to nil pointer, if the ingress configuration is empty.
This PR fixed this issue for the conversion webhook.


maximilien (approved):
/approve
/lgtm

PR #1029, opened 2022-04-07
Change int32 into *int32 for Replicas

### LINKED ISSUE #1028: Change int32 into *int32 for Replicas
Defining a field like Replica as int32 can leading to a situation that we have no idea it is unset or set to 0 by the user.
Suggest to replace with a pointer, so that nil means it is unset, and 0 means it is set to 0.

houshengbo:
Fixes #1028

**Release Note**

```release-note

```


maximilien (approved):
/approve
/lgtm

test/e2e-common.sh
```diff
@@ -49,7 +49,7 @@ release_eventing_yaml="$(mktemp)"
 
 readonly SERVING_ARTIFACTS=("serving" "serving-crds.yaml" "serving-core.yaml" "serving-hpa.yaml" "serving-post-install-jobs.yaml")
 readonly EVENTING_ARTIFACTS=("eventing" "eventing-crds.yaml" "eventing-core.yaml" "in-memory-channel.yaml" "mt-channel-broker.yaml"
-  "eventing-sugar-controller.yaml" "eventing-post-install.yaml")
```
csantanapr, line 52: Why `eventing-sugar-controller.yaml` removed?

pierDipi: it's now embedded in the eventing core controller, however, disabled by default.

PR #1047, opened 2022-04-22
Add the support of port and host configuration for gateways

### LINKED ISSUE #932: Suport of Knative-ingress-gateway and Knative-local-gateway overwriting
**Problem**
Knative-ingress-gateway and Knative-local-gateway are two key resources available in net-istio.
It turned out to be that these two resources are frequently modified based on the feedback of the community.
There was a debate about whether they are knative resources, whether they should be the scope of knative serving.

However, since how to change these two has been asked as one of the most frequently asked questions, could operator make it easier for users to configure them? Regardless they are knative resource or not, they are out of or within the scope, etc.

Link to an archived PR with issue: https://github.com/knative/serving-operator/pull/258

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):**
How many developer-days do you think this may take to resolve?

Add any other context about the feature request here.

houshengbo:
Fixes #932

## Proposed Changes
* It has been years since this feature was proposed for the first time. Although there have been strong debates in the communities, I finally believe it is a reasonable configuration option for gateways in net-istio. Istio configuration is not what knative covers, but net-istio as a valid network ingress plugin in knative operator, knative operator should be able to configure the gateways in net-istio.

**Release Note**

```release-note

```


maximilien (commented):
/ok-to-test

houshengbo:
I would like to give personal credit to https://github.com/eqqe, since he firstly opened the pr for the change.

maximilien (approved):
/approve
/lgtm

Issue #915, opened 2022-01-07
support add environment variable to conntainer


wu0407:
**Problem**
I want to add environment KUBERNETES_MIN_VERSION to container, but current version operator not have method to setting. 

**[Persona:](https://github.com/knative/eventing/blob/main/docs/personas.md)**
Which persona is this feature for?
System Operator

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):**
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

houshengbo:
@wu0407 Could you be more specific and explain more in details about the issue you have?
Could this be generalized into a broader issue like setting the env var for a container?

wu0407:
I running knative v1.1.0 in kubernetes v1.18,  need to set `KUBERNETES_MIN_VERSION` to overide required mini version.

houshengbo:
I suggest you NOT do that. Don't force to change the min version of Kubernetes.

wu0407:
I known that has some risk, but I want quick test or develop knative new version. I already check knative v1.1.0 is not depend on specific api that is in v1.20 but not in v1.18  before i do that.

onwsk8r:
What if I wanted to add an environment variable called `DEFAULT_JETSTREAM_URL` to the `jetstream-ch-dispatcher` deployment I get when setting `source.natss.enabled: true`?

PR #1083, opened 2022-05-26
Add support for net-* secret informer filtering

skonto:
## Proposed Changes

* Part of the work here: https://github.com/knative-sandbox/net-istio/commit/d5f9c0a00c0d4fd68208650fd84e89c33c1e1aa2
*  Introduces an annotation `knative.dev/enable-secret-informer-filtering-by-cert-uid` on Serving CR to trigger the related behavior. This can be done as follows: `knative.dev/enable-secret-informer-filtering-by-cert-uid`: `true`.
In the future is should be removed as the behavior will become the default.
Due to how [informers are consumed](https://github.com/knative/pkg/tree/main/injection#consuming-informers) this cannot reside in a config map as setup needs to happen early before sharedmain is called. Thus it is passed to net-istio via en env var.
Annotations are common on CRs to fine-tune behavior.
* A similar approach will be needed for net-kourier. The annotation will cover that too.

To test this follow instructions [here](https://gist.github.com/skonto/d051757ffa092bef8c6c0e841d49deb9).
/cc @houshengbo 
**Release Note**

```release-note
Users can enable informer filtering by decorating Serving CR with annotation `knative.dev/enable-secret-informer-filtering-by-cert-uid`:`true`.
```

### DIFF
--- a/pkg/reconciler/knativeserving/ingress/istio.go
+++ b/pkg/reconciler/knativeserving/ingress/istio.go
@@ -18,11 +18,15 @@ package ingress
 
 import (
 	"context"
+	"strconv"
 
 	mf "github.com/manifestival/manifestival"
 	"go.uber.org/zap"
 	istionetworkingv1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
 	"istio.io/client-go/pkg/clientset/versioned/scheme"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"knative.dev/operator/pkg/apis/operator/base"
 	"knative.dev/operator/pkg/apis/operator/v1beta1"
@@ -30,11 +34,20 @@ import (
 	"knative.dev/pkg/logging"
 )
 
+const enableSecretInformerFilteringByCertUIDAnno = "knative.dev/enable-secret-informer-filtering-by-cert-uid"
+
 var istioFilter = ingressFilter("istio")
+
+func init() {
+	_ = appsv1.AddToScheme(scheme.Scheme)
+}
 
 func istioTransformers(ctx context.Context, instance *v1beta1.KnativeServing) []mf.Transformer {
 	logger := logging.FromContext(ctx)
-	return []mf.Transformer{gatewayTransform(instance, logger)}
+	return []mf.Transformer{
+		gatewayTransform(instance, logger),
+		enableSecretInformerFilteringByCertUID(instance, logger),
+	}
 }
 
 func gatewayTransform(instance *servingv1beta1.KnativeServing, log *zap.SugaredLogger) mf.Transformer {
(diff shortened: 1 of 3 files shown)


pkg/reconciler/knativeserving/ingress/istio.go
```diff
@@ -18,23 +18,36 @@ package ingress
...
 	servingv1beta1 "knative.dev/operator/pkg/apis/operator/v1beta1"
 	"knative.dev/pkg/logging"
 )
 
+const enableSecretInformerFilteringByCertUIDAnno = "knative.dev/enable-secret-informer-filtering-by-cert-uid"
+
 var istioFilter = ingressFilter("istio")
 
+func init() {
+	_ = appsv1.AddToScheme(scheme.Scheme)
```
skonto, line 42: Needed for the deployment kind.

pkg/reconciler/knativeserving/ingress/istio.go
```diff
@@ -95,3 +108,34 @@ func updateIstioGateway(override *base.IstioGatewayOverride, gateway *istionetwo
...
+				}
+				deployment := &appsv1.Deployment{}
+				if err := scheme.Scheme.Convert(u, deployment, nil); err != nil {
+					return err
+				}
+				for i, c := range deployment.Spec.Template.Spec.Containers {
+					ptr := &c
+					if ptr.Name == "controller" {
+						ptr.Env = append(ptr.Env, corev1.EnvVar{
+							Name:  "ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID",
```
skonto, line 128: Don't want to bring in net-istio deps for the env var def only.

nak3:
Operator CR already supports `spec.deployments` to override `replicas`, `labels`, `nodeSelector` and so on - [docs link](https://knative.dev/development/install/operator/configuring-serving-cr/#override-the-resources).

So I wonder that it is better to support `spec.deployments.env` in general like:

```yaml
  deployments:
  - name: net-istio-controller
    env:
    - container: controller
      name: ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID:
      value: "true"
```

Actually net-kourier supports some env variables like https://github.com/knative-sandbox/net-kourier#setup-tls-certificate but operator cannot configure them.

skonto:
@nak3 sure :) that is an option if overriding env vars can be supported. One benefit of this PR is simpler UX.

pkg/reconciler/knativeserving/ingress/istio.go
houshengbo: I still don't think we need this init() function. In the transformer, when the `enable-secret-informer-filtering-by-cert-uid` is enabled, the deployment `net-istio-controller` will be changed and returned.
The list of resources in the manifest is changed, and applied in the reconciling loop, so it should be good for Kubernetes to pick up the changes.

pkg/reconciler/knativeserving/ingress/istio.go
```diff
@@ -18,23 +18,36 @@ package ingress
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"knative.dev/operator/pkg/apis/operator/base"
 	"knative.dev/operator/pkg/apis/operator/v1beta1"
 	servingv1beta1 "knative.dev/operator/pkg/apis/operator/v1beta1"
 	"knative.dev/pkg/logging"
 )
 
+const enableSecretInformerFilteringByCertUIDAnno = "knative.dev/enable-secret-informer-filtering-by-cert-uid"
```
houshengbo, line 37: Kubernetes operator's custom resources are able to configure the operands, by leverage the CR's spec. It is a bit strange to use the label to configure.

I suggest not bother the fields unless they are under `spec`.
You can add a boolean flag in the struct `type IstioIngressConfiguration struct` and change the serving CRD according. Let's configure everything with the CR's spec.

houshengbo:
It is good to think of supporting overriding env in general.
Let's see if there are other use cases. If @nak3 @skonto @dprotaso folks happen to know any, plz share them.

nak3:
> Let's see if there are other use cases. If @nak3 @skonto @dprotaso folks happen to know any, plz share them.

All controllers have `KUBERNETES_MIN_VERSION` env variables and some users want to change it. In fact, https://github.com/knative/operator/issues/915 was requested.
Some other env variables are also requested by https://github.com/knative/operator/issues/915#issuecomment-1065141629.

skonto:
I will work on a PR for the env support. From a UX perspective it might be better to have specific fields as the CR is the API users should use and not the env vars.

pkg/reconciler/knativeserving/ingress/istio.go
skonto: It is an annotation and it common practice to have such extra config info passed via annotations. I thought about that but it looked a bit intrusive to add in the CR since this looks an internal feature and it might be always on at some point.

houshengbo:
I suggest we close this PR for now, since we merge the one supporting evn vars, which can meet the requests for filtering resources. We can revisit this change, when the request for filtering gets more mature.

PR #1085, opened 2022-05-31
Support overriding env vars per container

skonto:
## Proposed Changes
Fixes: #915 

* Adds support for env overrides in deployments field
*  Discussed use cases in #1083 
* Sample cr for testing:
 ```
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  deployments:
    - name: controller
      env:
          - container: controller 
            envVars: 
              - name: METRICS_DOMAIN
                value: "knative.dev/skonto"
```
Output: 
```
    Environment:
      POD_NAME:                   controller-69f9f84d7-qwj28 (v1:metadata.name)
      SYSTEM_NAMESPACE:           knative-serving (v1:metadata.namespace)
      CONFIG_LOGGING_NAME:        config-logging
      CONFIG_OBSERVABILITY_NAME:  config-observability
      METRICS_DOMAIN:             knative.dev/skonto
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-j269x (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             True 
  ContainersReady   True 
  PodScheduled      True 
Volumes:
  kube-api-access-j269x:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
QoS Class:                   Burstable
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type    Reason     Age   From               Message
  ----    ------     ----  ----               -------
  Normal  Scheduled  35m   default-scheduler  Successfully assigned knative-serving/controller-69f9f84d7-qwj28 to minikube
  Normal  Pulling    35m   kubelet            Pulling image "gcr.io/knative-releases/knative.dev/serving/cmd/controller@sha256:8d84706d53adcf89c49687b4fade06261769b9f99257cb64d1758398f085b062"
  Normal  Pulled     35m   kubelet            Successfully pulled image "gcr.io/knative-releases/knative.dev/serving/cmd/controller@sha256:8d84706d53adcf89c49687b4fade06261769b9f99257cb64d1758398f085b062" in 9.203030199s
  Normal  Created    35m   kubelet            Created container controller
  Normal  Started    35m   kubelet            Started container controller

```

**Release Note**

```release-note
Serving CR supports env overriding per container via a new field in deployments obj.
```


pkg/apis/operator/base/common.go
```diff
@@ -291,6 +295,13 @@ type ResourceRequirementsOverride struct {
 	corev1.ResourceRequirements
 }
 
+type EnvRequirementsOverride struct {
+	// The container name
+	Container string `json:"container"`
+	// The desired EnvVarRequirements
+	EnvVars []corev1.EnvVar `json:"envVars,omitempty"`
```
skonto, line 302: I could use `corev1.EnvVar` to make it more redundant as with the resource overrides but I find it more compact to allow group multiple env vars per container name.

pkg/reconciler/common/deployments_override_test.go
```diff
@@ -118,6 +119,48 @@ func TestDeploymentsTransform(t *testing.T) {
 			expTemplateAnnotations: map[string]string{"cluster-autoscaler.kubernetes.io/safe-to-evict": "true"},
 			expReplicas:            10,
 		}},
+	}, {
+		name: "override env vars",
```
skonto, line 123: I can add more tests here eg. empty case.

houshengbo: @skonto Try to add test cases with more coverage. Otherwise it looks good to me. Wait for @nak3 to see if he has feedback.

pkg/reconciler/common/env_override.go
```diff
@@ -0,0 +1,54 @@
...
+func findEnvOverride(resources []base.EnvRequirementsOverride, name string) *base.EnvRequirementsOverride {
+	for _, override := range resources {
+		if override.Container == name {
+			return &override
+		}
+	}
+	return nil
+}
+
+func getEnv(containers []v1.Container, container string) []v1.EnvVar {
```
nak3, line 47: This function is used by unit test only (`deployments_override_test.go`). It should be moved to `deployments_override_test.go`?

nak3 (changes requested):
In my understanding, this implementation supports only "override". It does not support to "add" a new env value so users can change the env variable which is defined in the release manifest by default. Is this an expected design? I wonder that it is alright to support "add".

skonto:
> @nak3 In my understanding, this implementation supports only "override". It does not support to "add" a new env value so users can change only env variable which is defined in the release manifest by default. Is this an expected design? I wonder that it is alright to support "add".

That was my intention to change something existing legitimate env vars. Also if there are no env vars you can override with a new array of env vars. However, I think we should allow adding a new one. I will fix that.

skonto:
@nak3 @houshengbo updated the PR. Now env vars can be added and more test coverage is done. Note here that when env vars with the same name are defined are overwritten with the last defined taking precedence. In K8s when anv vars have the same name in the pod spec K8s validation reports a warning but the pod spec is applied and the last defined env var is applied. For more check [this](https://gist.github.com/skonto/789af5cfa7578e9e340e0e74fc59a1be) and the related test cases.

skonto:
`update-codegen.sh` fails on main branch too. Is this something that will be updated automatically @houshengbo so I can rebase later? Security request failed, will retry.

xfu83:
@houshengbo hi Vincent, can you help also cherry pick for the release-1.3? thx

PR #1160, opened 2022-08-11
Tolerate the case that the existing eventing version is not found

houshengbo:
Signed-off-by: Vincent Hou <shou@us.ibm.com>


maximilien (approved):
/approve
/lgtm

PR #1179, opened 2022-08-15
Change the order of resources and revert the webhook changes

houshengbo:


maximilien (approved):
I wonder if this "optimal" order changes with Knative releases? Or Kube?

/approve
/lgtm

PR #1182, opened 2022-08-16
Add the webhook for future CRD future conversion

houshengbo:
## Proposed Changes
* As Knative Operator starts to stabilize the APIs, we plan to release the v1 version. As the ground work, we need to make the conversion webhook ready.

**Release Note**

```release-note

```


maximilien (approved):
/approve
/lgtm

PR #1193, opened 2022-08-24
Prepare 1.7 release and update the deps and manifests

houshengbo:


test/resources/knativeserving.go
```diff
@@ -113,7 +113,7 @@ func getTestKSOperatorCRSpec() v1beta1.KnativeServingSpec {
 		CommonSpec: base.CommonSpec{
 			Config: base.ConfigMapData{
 				DefaultsConfigKey: {
-					"revision-timeout-seconds": "200",
+					"revision-timeout-seconds": "300",
```
dprotaso, line 116: Can you revert this when you bump to serving v1.7.1?

Issue #1200, opened 2022-09-09
webhook HPA attempt at scaling up apparently reverted by the operator


nak3:
**Describe the bug**

Running KnativeServing with default highAvailability setting (2 replicas by default) under a heavier load of creating ~300 ksvcs at the same time, I see an an attempt at the webhook HorizontalPodAutoscaler to scale up webhook replicas, but any new pods are immediately deleted

```
webhook-69cff6d4b9-2dhgd                                       2/2     Running       0          16h
webhook-69cff6d4b9-7gj9b                                       2/2     Running       0          15h
webhook-69cff6d4b9-fg8vj                                       1/2     Terminating   0          22s
webhook-69cff6d4b9-l9rt8                                       2/2     Terminating   0          7s
webhook-69cff6d4b9-sdnwr                                       2/2     Terminating   0          7s
```

**Expected behavior**

webhook should be scaled up via HPA.

**Additional note**
Thank you @maschmid for reporting this issue.

PR #1201, opened 2022-09-09
Do not change `spec.replicas` directory for the Deployment which has HPA

nak3:
When deployment has a HPA (e.g. acativator, webhook etc), the HPA should modifiy the `spec.replicas` in deployment.

However, operator's `spec.high-availability` also updates the `spec.replicas` in Deployment, so currently the replicas scaled up by HPA are terminated by the opreator due to the confliction between HPA's replica number and operator's replica number.

Operator already controlls HPA's minReplicas so this patch just skips the duplicated controll of the operator.

Fixes https://github.com/knative/operator/issues/1200

**Release Note**

```release-note
The bug which fails to scale up via HPA under the high load is fixed.
```

### DIFF
--- a/pkg/reconciler/common/ha.go
+++ b/pkg/reconciler/common/ha.go
@@ -30,6 +30,18 @@ func haUnSupported(obj base.KComponent) sets.String {
 	)
 }
 
+// When Deployment has HPA, the replicas should be controlled by HPA's minReplicas instead of operator.
+// Hence, skip changing the spec.replicas in deployment directory for these Deployments.
+func hasHorizontalPodAutoscaler(obj base.KComponent) sets.String {
+	return sets.NewString(
+		"webhook",
+		"activator",
+		"eventing-webhook",
+		"mt-broker-ingress",
+		"mt-broker-filter",
+	)
+}
+
 // HighAvailabilityTransform mutates configmaps and replicacounts of certain
 // controllers when HA control plane is specified.
 func HighAvailabilityTransform(obj base.KComponent, log *zap.SugaredLogger) mf.Transformer {
@@ -50,7 +62,7 @@ func HighAvailabilityTransform(obj base.KComponent, log *zap.SugaredLogger) mf.T
 		replicas := int64(*ha.Replicas)
 
 		// Transform deployments that support HA.
-		if u.GetKind() == "Deployment" && !haUnSupported(obj).Has(u.GetName()) {
+		if u.GetKind() == "Deployment" && !haUnSupported(obj).Has(u.GetName()) && !hasHorizontalPodAutoscaler(obj).Has(u.GetName()) {
 			if err := unstructured.SetNestedField(u.Object, replicas, "spec", "replicas"); err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


pkg/reconciler/common/ha.go
```diff
@@ -50,7 +62,7 @@ func HighAvailabilityTransform(obj base.KComponent, log *zap.SugaredLogger) mf.T
 		replicas := int64(*ha.Replicas)
 
 		// Transform deployments that support HA.
-		if u.GetKind() == "Deployment" && !haUnSupported(obj).Has(u.GetName()) {
+		if u.GetKind() == "Deployment" && !haUnSupported(obj).Has(u.GetName()) && !hasHorizontalPodAutoscaler(obj).Has(u.GetName()) {
```
skonto, line 65: Should we also respect the user setting by setting replicas as usual and also by `minReplicas` in the horizontalPodAutoscaler, an example is here?

nak3: Hmm.... I think https://github.com/openshift-knative/serverless-operator/blob/13833dae7e3f58294705cf2ad20ea6ae1573fccb/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go#L659 also has the same issue with https://github.com/knative/operator/issues/1200 .

The operator always sets the replicas based on `Spec.HighAvailability.Replicas` to Deployment so it conflicts when HPA updates the deployment during the high load.

nak3: To clarify, the issue happens the second case below.
#### Usual (no confliction happens :angel: ):
- operator sets replicas to `2` in deployment (based on `spec.HighAvailability.Replicas`)
- HPA sets replicas to `2` in deployment (based on `minReplicas`)

#### High load (confliction happens :fire:)
- operator sets replicas to `2` in deployment (based on `spec.HighAvailability.Replicas`)
- HPA sets replicas to `5` in deployment (based on the high load)

So, this PR changes to let operator skip changing the replicas for deployment which has HPA.

skonto: @pierDipi ^^^ fyi.
Ok. So when operator deployment informer triggers a reconciliation we should skip applying setting replicas favoring hpa for the deployments that have an hpa config. 
Should we make sure though that the minScale for hpa is set to spec.HighAvailability.Replicas in all scenarios?

nak3: I think current logic for hpa is fine.

pkg/reconciler/common/ha.go
```diff
@@ -30,6 +30,18 @@ func haUnSupported(obj base.KComponent) sets.String {
...
+// Hence, skip changing the spec.replicas in deployment directory for these Deployments.
+func hasHorizontalPodAutoscaler(obj base.KComponent) sets.String {
+	return sets.NewString(
+		"webhook",
+		"activator",
+		"eventing-webhook",
+		"mt-broker-ingress",
+		"mt-broker-filter",
+	)
+}
```
pierDipi, line 43: I think downstream we remove some HPA config, so I think we should not hardcode the list

nak3: Do you have any other good/simple way to solve it without the hardcode?
While I also think that the hardcode the list is not good, I wonder that upstream does not need to consider about downstream's their own customization".

nak3: Anyway I am not familiar with eventing so I dropped the eventing components from the list.

houshengbo:
/lgtm
/approve
Thx for fixing this issue!!!!!!!

PR #1213, opened 2022-09-27
Support overriding env vars per container for 1.3.x

xfu83:
## Proposed Changes

* Cherry picked the PR https://github.com/knative/operator/pull/1085 to 1.3.x release branch, which make operator support env vars per container so we can override `KUBERNETES_MIN_VERSION`

**Release Note**

```release-note
Serving CR supports env overriding per container via a new field in deployments obj.
```


houshengbo:
1.3 cannot accept this change as new patch release, because it changes the CRD schema, which will break the upgrade and downgrade of knative operator. @xfu83

xfu83:
@houshengbo I see. Close this since it's superseded by https://github.com/knative/operator/pull/1220

PR #1228, opened 2022-09-29
Add the code to safeguard the nil pointer issue

houshengbo:
Fixes: #1211


pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -85,6 +85,10 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1beta1.Knative
 		return nil
 	}
 
+	if manifest == nil {
+		return nil
```
lionelvillard, line 89: what about logging to indicate there is no manifest?

maximilien: +1

maximilien (approved):
Please add logging for LGTM

/approve

PR #1231, opened 2022-10-03
Refactor the ingress service for istio

### LINKED ISSUE #1206: knative-local-gateway service selectors assume istio defaults
**Describe the bug**
If you use the istio ingress, but configure it to use a different selector than the default "`istio: ingressgateway`", then the `knative-local-gateway` k8s service created by the operator will still be set with its selectors set to `istio: ingressgateway` so will fail to route traffic. As a result, `DomainMapping`s and services talking to other private services in the cluster will not work.

**Expected behavior**
The `knative-local-gateway` k8s service is configured with selectors that match istio-ingress pods.

**To Reproduce**
Install an istio ingress gateway to the default namespace `istio-system` with its labels configured to something other than the default `istio: ingressgateway`. We'll use `istio: ingress` instead for example.

(You could use the istio operator, adjusting the labels in `ingress.yaml`, see [istio docs](https://istio.io/latest/docs/setup/additional-setup/gateway/#deploying-a-gateway)

Install knative using the operator, including these values in the `knativeserving` CRD:
```
spec:
  config:
    istio:
      gateway.knative-serving.knative-ingress-gateway: istio-ingress.istio-system.svc.cluster.local
      local-gateway.knative-serving.knative-local-gateway: knative-local-gateway.istio-system.svc.cluster.local
  ingress:
    istio:
      enabled: true
      knative-ingress-gateway:
        selector:
          istio: ingress
      knative-local-gateway:
        selector:
          istio: ingress
```

**Knative release version**
1.5

**Additional context**
I propose solving this by setting the selectors to match whatever is currently found on the public k8s service. I think its safe to assume the traffic should always be forwarded to the same istio-ingress pods that handle the public knative traffic, because I don't believe installing knative to use more than one istio gateway is supported.

I like this solution better than one which might at first appear more obvious: using the values from the `knativeserving` CRD under `ingress.istio.knative-local-gateway.selector`. This is because these are supposed to control the selectors on the istio Gateway resource, not the k8s service, and if we were to change that to set the selectors on the services too, then one would expect changing `ingress.istio.knative-ingress-gateway.selector` would adjust the associated public k8s service too. But the public service doesn't have this bug -- it's not included in the manifest or touched at all by this operator.

Past related PRs & issues researched to reach proposed solution:
* https://github.com/knative/operator/pull/590
* https://github.com/knative/operator/issues/831 (the same bug really, didn't get fixed, was unceremoniously closed by a bot)

houshengbo:
Fixes #1206

### DIFF
--- a/pkg/reconciler/knativeserving/ingress/ingress_service.go
+++ b/pkg/reconciler/knativeserving/ingress/ingress_service.go
@@ -20,6 +20,10 @@ import (
 	"fmt"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"knative.dev/operator/pkg/apis/operator/base"
+
 	mf "github.com/manifestival/manifestival"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"knative.dev/operator/pkg/apis/operator/v1beta1"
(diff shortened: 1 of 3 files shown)


mbaynton:
I've confirmed that this works for me wrt. my issue reported in #1206. Thanks much @houshengbo!

Not sure if I'm bestowed with powers for this tag to matter, but /lgtm

maximilien (approved):
/approve
/lgtm

PR #1242, opened 2022-10-07
Update the fetcher and manfests for kafka sources

### LINKED ISSUE #1216: Configuring Knative to use a local gateway in another namespace
(sorry, cross-post with knative-sandbox/net-istio#985.  I posted there before realising this is likely an operator issue)

**Describe the bug**
I am trying to use Knative Serving via the Knative Operator to make `ksvcs` available via a local gateway (not a service mesh).  The `Gateway` I have is shared with other applications, and the workload pods for that gateway are in a separate namespace.  I'm having trouble getting the local side of the gateway to work properly, specifically with the `istio-proxy` deployment deployed in a different namespace from serving/any ksvc.  I wonder if there's a missing/misconfigured `service` that the net istio controller needs to generate.  

**Expected behavior**
I'd expect the operator to configure Gateways and services that allow for in-cluster communication.

**To Reproduce**
I have:
* deployed istio in `namespace-1`, including 
  * an `image: istio/proxy` deployment working as a gateway deployed with selector `istio: ingressgateway` 
  * an external `Gateway` pointing to the above proxy called `my-gateway`
  * a `svc` called `istio-ingressgateway-workload` that points at the above proxy via a selector
* deployed the Knative Operator
* created a `KnativeServing` object in `namespace/knative-serving` using:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: knative-serving
---
apiVersion: v1
kind: Namespace
metadata:
  name: knative-serving
---
kind: KnativeServing
metadata:
  name: knative-serving-app
  namespace: knative-serving
spec:
  version: "1.7.1"
  config:
    istio:
      gateway.namespace-1.my-gateway: "istio-ingressgateway-workload.namespace-1.svc.cluster.local"
      local-gateway.knative-serving.knative-local-gateway: "knative-local-gateway.knative-serving.svc.cluster.local"
    domain:
      10.64.140.43.nip.io: ""
```

Once I have these, I deploy a cloudevents player with:
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: cloudevents-player-1
spec:
  template:
    metadata:
      annotations:
        autoscaling.knative.dev/minScale: "1"
    spec:
      containers:
        - image: quay.io/ruben/cloudevents-player:0.2
```

Once up, I can connect to the `ksvc` through the external gateway no problem using:
```sh
curl -v http://cloudevents-player-1.namespace-1.10.64.140.43.nip.io -H "Content-Type: application/json" -H "Ce-Id: foo" -H "Ce-Specversion: 1.0" -H "Ce-Type: dev.example.events" -H "Ce-Source: curl-source" -d '{"msg":"Can you hear me?"}'
```

which yields a `202 Accepted`.  However, if I try the local ingress, I get connection refused:
```sh
curl -v http://cloudevents-player-1.namespace-1.svc.cluster.local -H "Content-Type: application/json" -H "Ce-Id: foo" -H "Ce-Specversion: 1.0" -H "Ce-Type: dev.example.events" -H "Ce-Source: curl-source" -d '{"msg":"Can you hear me?"}'
```

which yields a `Failed to connect ... port 80: Connection refused`.

**Knative release version**
1.1 and 1.7

**Additional context**
I think this is because the `svc/knative-local-gateway` which is generated by the istio controller tries to point to the proxy via a selector, but the proxy is in a different namespace.  We see the service (truncated a bit):
```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
  name: knative-local-gateway
  namespace: knative-serving
spec:
  ports:
  - name: http2
    port: 80
    protocol: TCP
    targetPort: 8081
  selector:
    istio: ingressgateway
  type: ClusterIP
```

which is in `namespace/knative-serving` and tries to select pods by `selector: {istio: ingressgateway}`.  Problem is that there is no `istio: ingressgateway` in `namespace/knative-serving`, it is instead deployed in `namespace/namespace-1`.  

Am I configuring this incorrectly?  I don't see any other settings that might change this.  Maybe the intent is for me to point the local gateway at my own service, but then I don't understand why knative would create a `svc/knative-local-gateway` at all.  The existing `svc/istio-ingressgateway-workload` that I have would be unsuitable as it is forwarding port 80 to 8080 (eg: the external route through the gateway).  I'd need another svc to pass to port 8081 (like the `svc/knative-local-gateway` does).

If everything is configured correctly, I think that `svc/knative-local-gateway` needs to either be created in the namespace of the istio proxy pod, or be an `ExternalName` svc that resolves to a service in the istio proxy pod's namespace and points at the istio proxy pod.

houshengbo:
Fixes #1216


maximilien (approved):
/approve
/lgtm

Issue #762, opened 2021-09-09
`spec.deployments.replicas` does not take care about HorizontalPodAutoscaler


nak3:
We can overrides `spec.deployments.replicas` as:

```
apiVersion: operator.knative.dev/v1alpha1
kind: KnativeServing
metadata:
  name: ks
  namespace: knative-serving
spec:
  deployments:
  - name: webhook
    replicas: 3
```

But it does not change the values in `HorizontalPodAutoscaler`:

https://github.com/knative/operator/blob/f4ce4ff51a7fe3f6941073c7733a2e83f8b7faf1/pkg/reconciler/common/deployments_override.go#L86-L90

Some deployments (e.g. webhook, activator) deploys HPA so we need to change the values.

https://github.com/knative/operator/blob/f4ce4ff51a7fe3f6941073c7733a2e83f8b7faf1/cmd/operator/kodata/knative-serving/0.25.1/2-serving-core.yaml#L5435

matzew:
@nak3 we have some code updating the HPA `minReplica` here: https://github.com/knative/operator/blob/main/pkg/reconciler/common/ha.go#L76-L89

I think we should do something similar on the `deployments_override.go` file, to check for both: `min` and `max`...

maschmid:
I think we need an explicit option like
```
spec:
  horizontalPodAutoscalers:
  - name: foo
    minReplicas: ...
    maxReplicas: ...
```

+ a warning if one attempts to set spec.deployments[].replicas for something the operator does deploy an HPA...

matzew:
or that as a vehicle to make more options directly configurable, w/o guessing.

Regardless, it would also would relate to the `ha.go` file

nak3:
Yeah, `maxReplicas` has a bug in not only `spec.deployments.replicas` but also `spec.ha` (`ha.go`).
I think we can apply the same fix with https://github.com/knative/operator/pull/748 to `spec.deployments.replicas`.

houshengbo:
@nak3 Can we close this issue for now?

richardvflux:
not sure what frozen means, its still not fixed :-(

dprotaso:
Frozen means the bot won't close this issue.

PR #1245, opened 2022-10-11
Deprecate HA field

matzew:

Fixes #

## Proposed Changes

* Deprecate HA in fav. of `WorkloadOverride`
*
*

**Release Note**

```release-note
The high-availability field is deprecated and will be removed in a future version. Please use WorkloadOverride to specifiy replica number
```

### DIFF
--- a/pkg/apis/operator/base/common.go
+++ b/pkg/apis/operator/base/common.go
@@ -163,6 +163,7 @@ type CommonSpec struct {
 	AdditionalManifests []Manifest `json:"additionalManifests,omitempty"`
 
 	// HighAvailability allows specification of HA control plane.
+	// Deprecated: use WorkloadOverride to specify the replica on a per deployment base.
 	// +optional
 	HighAvailability *HighAvailability `json:"high-availability,omitempty"`
 


nak3:
I think we can also keep the HA field with `WorkloadOverride` i.e. - HA field can apply all deployments but users can also override the replicas via `WorkloadOverride` (I guess it is current behavior). I cannot deny that setting `replica` for every single deployments/statefulsets is so hard.

With said, I don't have any strong option for keep HA field so I'm fine for the deprecation and would like to see others' options.

matzew:
That's a good point.

HA could be use to apply for all deployments, as is, but workload wins as an override?

Or is `workloads` the way to go, b/c you can scale per individual workload (deployment/sts)

nak3:
> HA could be use to apply for all deployments, as is, but workload wins as an override?

Yes, I think it is the current behavior. The transformer is applied in this order  
https://github.com/knative/operator/blob/d520f2e40a37b677aa847477613702193bea8880/pkg/reconciler/common/transformers.go#L28-L39

But again I am still not sure if we just simply use `workloads` only or not.

nak3:
Right, there is no easy way to configure all deployments with HA=N.

But please just let me also explain about the motivation for the deprecation. (I am on neither side :sweat_smile: )
The current code is complicated and a little bit hard to maintain due to the two HA settings.
* For example, I think https://github.com/knative/operator/issues/762 (confliction with HPA's max/min replicas) is not fixed yet because of the HA setting's complication.
* We need to maintain [`haUnSupported` `hasHorizontalPodAutoscaler` ](https://github.com/knative/operator/blob/main/pkg/reconciler/common/ha.go#L28-L41) by the fixed name.

The [2021's survey](https://docs.google.com/presentation/d/1yGRsNoUlRc4F8L5CBcqBO-fPziR30o4vIWCW-tkWL_M/edit?resourcekey=0-ILz56oFJ7cd55I5gAhi5Xg#slide=id.gecc00d770b_1_2) showed that the configuration is used by 19%. But hmm I'm not sure if we should see the number is high or low :sweat_smile:

PR #1246, opened 2022-10-12
Add workloads and deprecate deployments 

### LINKED ISSUE #1155: Rename `deployments` field to `components` (or something else more generic)
**Problem**

The current approach of defining overrides for Knative components is to define a field, [called `deployments`](https://knative.dev/docs/install/operator/configuring-serving-cr/#override-system-deployments) like the following:
```yaml
spec:
  deployments:
  - name: controller
    resources:
    - container: controller
      requests:
        cpu: 300m
        memory: 100Mi
      limits:
        cpu: 1000m
        memory: 250Mi
```

While most of the components in eventing and serving are using deployments, some components use `StatefulSet`s and to have consistency across different components (even components currently unsupported by the operator) having 2 fields one for `deployments` and one for `statefulsets` will make the API harder to use.

The main reason to switch to a more generic name is that all fields in [`DeploymentOverride`](https://github.com/knative/operator/blob/b879b3e16c2391c47dc056ac9eb57e86dce9bebd/pkg/apis/operator/base/common.go#L246-L282) are applicable to `StatefulSet`s or `DeamonSet`s and users shouldn't necessarily understand which component is of which kind to use it since the override configurations will work trasparently. 

The new field `components` might be added in v1beta1 as an additional field and then once we move to v1 we can remove the `deployments` field.

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.
- Rename `deployments` field to `components` (or something else more generic than deployments)
- Rename  [`DeploymentOverride`](https://github.com/knative/operator/blob/b879b3e16c2391c47dc056ac9eb57e86dce9bebd/pkg/apis/operator/base/common.go#L246-L282) struct to `ComponentOverride`.

**Time Estimate (optional):**
How many developer-days do you think this may take to resolve?
5

Add any other context about the feature request here.

pierDipi:
As per title.

Fixes #1155 

## Proposed Changes

* Add `spec.workloads`
* Deprecated `spec.deployments`

**Release Note**

```release-note
knativeeventing.spec.deployments and knativeserving.spec.deployments is deprecated, use knativeeventing.spec.workloads and knativeserving.spec.workloads instead.
```

TODO: change documentation in knative/docs


houshengbo:
Per my comment here: https://github.com/knative/operator/issues/1155#issuecomment-1271740272
I will say do not deprecate the spec.deployment, at least in the current v1beta1 CRDs.

Once we have a new version of CRDs, we can deprecate the spec.deployment.

pierDipi:
cc @houshengbo

matzew:
I'd say we add the deprecated warning, as a vehicle to communicate that things are changing, via the code

pkg/reconciler/common/components_override.go
```diff
@@ -0,0 +1,153 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes/scheme"
+
+	"knative.dev/operator/pkg/apis/operator/base"
+)
+
+// ComponentsTransform transforms deployments based on the configuration in `spec.deployment`.
+func ComponentsTransform(obj base.KComponent, log *zap.SugaredLogger) mf.Transformer {
```
matzew, line 32: can't we just pass in the `[]ComponentOverride`, instead of the `KComponent`, so this is usable on other operators that are _not_ implementing the `KComponent` contract?

pkg/reconciler/common/transformers.go
```diff
@@ -32,7 +33,7 @@ func transformers(ctx context.Context, obj base.KComponent) []mf.Transformer {
 		ImageTransform(obj.GetSpec().GetRegistry(), logger),
 		ConfigMapTransform(obj.GetSpec().GetConfig(), logger),
 		ResourceRequirementsTransform(obj, logger),
-		DeploymentsTransform(obj, logger),
+		ComponentsTransform(obj, logger),
```
matzew, line 36: and here we would to than `obj.GetSpec().GetComponentOverride()` ?

pkg/reconciler/common/components_override.go
pierDipi: +1

pierDipi:
@matzew done!

matzew (approved):
/lgtm
/approve

/hold
for others

As said: there is no harm in deprecating some API on this (`v1b1`) version, it will stay - it is just deprecated. It will be removed by a future release, e.g. when we bump our APIs to `v1`.

pkg/apis/operator/base/common.go
```diff
@@ -137,15 +137,20 @@ type CommonSpec struct {
...
 
+	// DEPRECATED. Use components
 	// DeploymentOverride overrides Deployment configurations such as resources and replicas.
 	// +optional
-	DeploymentOverride []DeploymentOverride `json:"deployments,omitempty"`
+	DeploymentOverride []WorkloadOverride `json:"deployments,omitempty"`
+
+	// ComponentsOverride overrides workloads configurations such as resources and replicas.
+	// +optional
+	ComponentsOverride []WorkloadOverride `json:"workloads,omitempty"`
```
skonto, line 147: I guess we can have WorkloadsOverride in the future?

skonto:
LGTM

pkg/apis/operator/base/common.go
pierDipi: I will rename it now to `Workloads`

houshengbo:
This PR needs to add more test coverage.

matzew (approved):
/lgtm

thanks for the updates on the tests, @pierDipi 

feel free to unhold

PR #1247, opened 2022-10-12
Support deployments/workloads probe overrides

skonto:
## Proposed Changes

* Customers are facing slow environments and need to adjust timeouts and other settings in probes.
* Some settings can be adjusted on the fly via changing the deployments eg. a deployment uses the probe defaults and given that the operator will not touch those (manifest uses apply) it does the trick. However this is a hack and we need a proper way to change probes and cover cases where reconciliation will touch the settings.

**Release Note**

```release-note
Support is added for readiness, liveness probe overrides.
```


skonto:
@houshengbo @nak3

pkg/apis/operator/base/common.go
```diff
@@ -279,6 +279,14 @@ type DeploymentOverride struct {
...
 	// +optional
 	Env []EnvRequirementsOverride `json:"env,omitempty"`
+
+	// ReadinessProbes overrides readiness probes for the containers.
+	// +optional
+	ReadinessProbes []ProbesRequirementsOverride `json:"readinessProbes,omitempty"`
+
+	// LivenessProbes overrides liveness probes for the containers.
+	// +optional
+	LivenessProbes []ProbesRequirementsOverride `json:"livenessProbes,omitempty"`
```
matzew, line 289: very valid to have these as overrides too! Good catch!

matzew:
Let's have a knative/docs issue as well, so we describe this valuable feature!

matzew:
LGTM

skonto:
I will create a doc PR for this too once it is approved. @houshengbo gentle ping.

pkg/apis/operator/base/common.go
```diff
@@ -323,6 +331,14 @@ type EnvRequirementsOverride struct {
 	EnvVars []corev1.EnvVar `json:"envVars,omitempty"`
 }
 
+// ProbesRequirementsOverride enables the user to override any container's env vars.
+type ProbesRequirementsOverride struct {
+	// The container name
+	Container string `json:"container"`
+	// The desired ProbeRequirements
+	Probe corev1.Probe `json:"probe,omitempty"`
```
pierDipi, line 339: Should we prevent setting the [`ProbeHandler` field](https://github.com/skonto/operator/blob/23892c03b5a5caeb2e41691a4bff41248903dd73/vendor/k8s.io/api/core/v1/types.go#L2462) and allow only these fields https://github.com/skonto/operator/blob/23892c03b5a5caeb2e41691a4bff41248903dd73/vendor/k8s.io/api/core/v1/types.go#L2194-L2226.

pkg/apis/operator/base/common.go
```diff
@@ -323,6 +331,14 @@ type EnvRequirementsOverride struct {
...
+}
```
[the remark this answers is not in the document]
matzew: > For v1, a more podspec-like structure (with a subset of fields) would be easier to work with.

yeah, currently is hard...

houshengbo:
Do we really need to all fields under `livenessProbes`???
If it is just a few fileds, like TimeoutSeconds, InitialDelaySeconds, etc, we do not need to add all the other fields. Just pick the ones related to the `*** Seconds`, that are necessary.

pkg/apis/operator/base/common.go
skonto: I didnt want to limit people on how they can use this but we can.

pkg/apis/operator/base/common.go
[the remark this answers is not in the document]
skonto: Yes I would prefer a pod spec per workload to apply as an override.

skonto:
> If it is just a few fileds, like TimeoutSeconds, InitialDelaySeconds, etc, we do not need to add all the other fields. Just pick the ones related to the *** Seconds, that are necessary.

Anything that is timeout related is useful as a starting point.

pkg/apis/operator/base/common.go
houshengbo: Let's not assume the options for users. If we plan to add some, what is the use case? Otherwise, limit to what is necessary. Drop ProbeHandler at least.

houshengbo: @skonto

houshengbo:
Just keep 
```
InitialDelaySeconds int32 `json:"initialDelaySeconds,omitempty" protobuf:"varint,2,opt,name=initialDelaySeconds"`
	
	TimeoutSeconds int32 `json:"timeoutSeconds,omitempty" protobuf:"varint,3,opt,name=timeoutSeconds"`
	
	PeriodSeconds int32 `json:"periodSeconds,omitempty" protobuf:"varint,4,opt,name=periodSeconds"`
	
	SuccessThreshold int32 `json:"successThreshold,omitempty" protobuf:"varint,5,opt,name=successThreshold"`
	
	FailureThreshold int32 `json:"failureThreshold,omitempty" protobuf:"varint,6,opt,name=failureThreshold"`
	
	TerminationGracePeriodSeconds
```
in the CRDs. Remove the others. @skonto

skonto:
@houshengbo kept those six fields. Adapted the PR.

houshengbo:
@skonto Plz do another rebase. I will merge this pr.

skonto:
@houshengbo sorry missed it I was off. Let's see if tests pass.

skonto:
fixed crds /unhold

matzew (approved):
/lgtm
/approve

/hold
feel free to unhold

PR #1297, opened 2022-12-09
Patch CVE-2022-41717

evankanderson:
Update go.mod to force a rebuild for CVE-2022-41717


evankanderson:
It looks like I need to run update_codegen, so I'll do that instead.

/close

PR #1324, opened 2023-01-24
Take workload overrides into account for setting `SINK_BINDING_SELECTION_MODE` env var

creydr:
Currently `SinkBindingSelectionModeTransform` sets the `SINK_BINDING_SELECTION_MODE` only based on the `.spec.sinkBindingSelectionMode` field.
In case `SINK_BINDING_SELECTION_MODE` is set the workload override of the eventing-webhook, this fields is ignored even when `.spec.sinkBindingSelectionMode` is set ([SRVKE-1360](https://issues.redhat.com/browse/SRVKE-1360)).

This PR addresses it and takes the `SINK_BINDING_SELECTION_MODE` env var from the workload override of the eventing-webhook into account, if `.spec.sinkBindingSelectionMode` is not set.

## Proposed Changes

* :bug: Take the `SINK_BINDING_SELECTION_MODE` env var from the workload override of the eventing-webhook into account, if `.spec.sinkBindingSelectionMode` is not set.

### DIFF
--- a/pkg/reconciler/knativeeventing/common/sinkbindingselectionmode.go
+++ b/pkg/reconciler/knativeeventing/common/sinkbindingselectionmode.go
@@ -79,7 +79,8 @@ func SinkBindingSelectionModeTransform(instance *eventingv1beta1.KnativeEventing
 }
 
 func sinkBindingSelectionModeFromWorkloadOverrides(instance *eventingv1beta1.KnativeEventing) string {
-	for _, workloadOverride := range instance.Spec.Workloads {
+	overrides := append(instance.Spec.Workloads, instance.Spec.DeploymentOverride...)
+	for _, workloadOverride := range overrides {
 		if workloadOverride.Name == "eventing-webhook" {
 			for _, envRequirement := range workloadOverride.Env {
 				if envRequirement.Container == "eventing-webhook" {
--- a/pkg/reconciler/knativeeventing/common/sinkbindingselectionmode.go
+++ b/pkg/reconciler/knativeeventing/common/sinkbindingselectionmode.go
@@ -42,7 +42,11 @@ func SinkBindingSelectionModeTransform(instance *eventingv1beta1.KnativeEventing
 
 			sinkBindingSelectionMode := instance.Spec.SinkBindingSelectionMode
 			if sinkBindingSelectionMode == "" {
-				sinkBindingSelectionMode = "exclusion"
+				if smFromWorkloadOverrides := sinkBindingSelectionModeFromWorkloadOverrides(instance); smFromWorkloadOverrides != "" {
+					sinkBindingSelectionMode = smFromWorkloadOverrides
+				} else {
+					sinkBindingSelectionMode = "exclusion"
+				}
 			}
 
 			for i := range deployment.Spec.Template.Spec.Containers {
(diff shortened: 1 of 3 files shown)


pkg/reconciler/knativeeventing/common/sinkbindingselectionmode.go
```diff
@@ -73,3 +77,21 @@ func SinkBindingSelectionModeTransform(instance *eventingv1beta1.KnativeEventing
 		return nil
 	}
 }
+
+func sinkBindingSelectionModeFromWorkloadOverrides(instance *eventingv1beta1.KnativeEventing) string {
+	for _, workloadOverride := range instance.Spec.Workloads {
+		if workloadOverride.Name == "eventing-webhook" {
```
houshengbo, line 83: Do you really need this configuration for `eventing-webhook` only?

pkg/reconciler/knativeeventing/common/sinkbindingselectionmode.go
```diff
@@ -73,3 +77,21 @@ func SinkBindingSelectionModeTransform(instance *eventingv1beta1.KnativeEventing
...
+			for _, envRequirement := range workloadOverride.Env {
+				if envRequirement.Container == "eventing-webhook" {
```
houshengbo, line 85: Do we have to pin the contain name to `eventing-webhook` only?

pkg/reconciler/knativeeventing/common/sinkbindingselectionmode.go
creydr: As I understood, only the eventing-webhook deployment has this env var (in this it is relevant). But that's probably something @pierDipi can say for sure.

pkg/reconciler/knativeeventing/common/sinkbindingselectionmode.go
creydr: That's a good point. In the eventing deployments for the eventing-webhook, I saw that the container is named that way (https://github.com/knative/eventing/blob/466d123181f39f90a5f38583445910ffe61cbb0d/config/core/deployments/webhook.yaml#L53-L91). But it probably doesn't hurt to check this on all containers.

pkg/reconciler/knativeeventing/common/sinkbindingselectionmode.go
matzew: yeah, that's from the webhook, see: https://github.com/knative/eventing/search?q=SINK_BINDING_SELECTION_MODE


That said, we do however override the actual upstream default on our own opertor's extension: 
https://github.com/openshift-knative/serverless-operator/blob/main/openshift-knative-operator/pkg/eventing/extension.go#L66-L69


Could we have some test for that? (there is some extension code and tests in this repo)

houshengbo: This change only enables the support in "instance.Spec.Workloads".
The resource "eventing-webhook" is a deployment. 
What if the user configure the env var in "instance.Spec.Deployments"?

creydr: :+1: addressed in 681c3ad0

PR #1338, opened 2023-02-13
Add certs secret name read from env

hhk7734:
## Proposed Changes

* update deployments yaml files

**Release Note**

```release-note

```

### DIFF
--- a/vendor/knative.dev/pkg/webhook/env.go
+++ b/vendor/knative.dev/pkg/webhook/env.go
@@ -22,10 +22,14 @@ import (
 	"strconv"
 )
 
-const portEnvKey = "WEBHOOK_PORT"
+const (
+	portEnvKey = "WEBHOOK_PORT"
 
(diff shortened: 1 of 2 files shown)


vendor/knative.dev/pkg/webhook/env.go
```diff
@@ -22,10 +22,14 @@ import (
 	"strconv"
 )
 
-const portEnvKey = "WEBHOOK_PORT"
+const (
```
houshengbo, line 25: Have you manually changed this file: vendor/knative.dev/pkg/webhook/env.go?
It is a dependency of this repository. You shouldn't change it here at all.
If you propose a change in knative.dev/pkg, please submit your change in knative.dev/pkg first, then update the dependencies here in this repository.

hhk7734: I didn't realized that the pkg repository exist.
I will submit this PR to knative.dev/pkg first.

Thanks.

PR #1343, opened 2023-02-21
Add Kourier LoadBalancerIP Config

### LINKED ISSUE #1290: Specify Kourier Load Balancer IP
**Problem**
There's no declarative way of specifying the Load Balancer IP that Kourier should use.

System Operator

**Exit Criteria**
There should be a way of specifying Kourier's Load Balancer IP using the Operator.
https://github.com/knative-sandbox/net-kourier/issues/747

theomessin:
Fixes #1290

## Proposed Changes

* Adds new `service-load-balancer-ip` config for the Kourier Gateway.
* This sets Kourier's Service `loadBalancerIP` to the configured value.
* Several Cloud Providers use this setting to specify the IP address that a service gets assigned:

> Some cloud providers allow you to specify the loadBalancerIP. In those cases, the load-balancer is created with the user-specified loadBalancerIP. If the loadBalancerIP field is not specified, the loadBalancer is set up with an ephemeral IP address. If you specify a loadBalancerIP but your cloud provider does not support the feature, the loadbalancerIP field that you set is ignored.

**Release Note**

```release-note
NONE
```


theomessin:
Example config after this Pull Request is merged which would automagically get Knative Serving working with Auto-TLS in Google Kubernetes Engine as long as the set IP address has been created in GCP.

```yaml
---
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  ingress:
    kourier:
      enabled: true
      service-loadBalancerIP: 1.2.3.4
  config:
    domain:
      "example.com": ""
    network:
      auto-tls: Enabled
      certificate-class: net-http01.certificate.networking.knative.dev
      http-protocol: Redirected
      ingress-class: kourier.ingress.networking.knative.dev
  additionalManifests:
    - URL: https://github.com/knative/net-http01/releases/download/knative-v1.9.0/release.yaml
```

nak3 (changes requested):
Thank you for this PR @theomessin 

But I realized that the `LoadBalancerIP` field has been deprecated in k8s 1.24. Is it possible to ask your cloud provider to support the annotation as suggested in the change log?

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#deprecation
> Deprecated Service.Spec.LoadBalancerIP. This field was under-specified and its meaning varies across implementations. As of Kubernetes v1.24, users are encouraged to use implementation-specific annotations when available. This field may be removed in a future API version.

theomessin:
Hi @nak3. Yes, it has been marked as deprecated. We're using Google Kubernetes Engine and as far as I can tell they haven't implemented any alternatives as of yet. They still recommend we use `loadBalancerIP`. I'd say the best thing to do is still add this in the operator.

pkg/reconciler/knativeserving/ingress/kourier.go
```diff
@@ -107,6 +108,28 @@ func configureGWServiceType(instance *v1beta1.KnativeServing) mf.Transformer {
 	}
 }
 
+// configureGWServiceLoadBalancerIP configures Kourier GW's service loadBalancerIP.
+func configureGWServiceLoadBalancerIP(instance *v1beta1.KnativeServing) mf.Transformer {
```
nak3, line 112: As per https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/
> Only applies to Service Type: LoadBalancer.

The `ServiceLoadBalancerIP` applies to service type `LoadBalancer` only and you also need to use it for `LoadBalancer`, right? Then, I would the add `if` condition in `configureGWServiceType()` to insert the logic to apply `ServiceLoadBalancerIP` like:

```go
                        case v1.ServiceTypeClusterIP, v1.ServiceTypeNodePort, v1.ServiceTypeLoadBalancer:
                                svc.Spec.Type = serviceType
                                if serviceType == v1.ServiceTypeNodePort && instance.Spec.Ingress.Kourier.ServiceLoadBalancerIP != ""{
                                        ...
                                }
```

Probably we should rename the function `configureGWServiceType()`  with `configureGWService()` then.

config/crd/bases/operator.knative.dev_knativeservings.yaml
```diff
@@ -2299,6 +2299,8 @@ spec:
                         type: boolean
                       service-type:
                         type: string
+                      service-loadBalancerIP:
```
nak3, line 2302: `loadBalancerIP` is same style with k8s's but camel case and kebab case mixed.
@houshengbo Do you have any suggestion for the config name - like `service-loadBalancerIP`, `service-load-balancer-ip` or somethinge else?

theomessin: I'm happy to change this to `service-load-balancer-ip` if that's better. Let me know.

pkg/reconciler/knativeserving/ingress/kourier.go
theomessin: It's not that simple. The default type for the `kourier` Service is `LoadBalancer`
https://github.com/knative/operator/blob/a3758e7915b60ea859cbc4951101708d92097302/cmd/operator/kodata/ingress/1.10/kourier/kourier.yaml#L576

If we do this, we'd have to change this early return check because a user might set `service-loadBalancerIP` but not `service-type`.
https://github.com/knative/operator/blob/037957ad5d3ccd22b12826a0a17f22b122d78a32/pkg/reconciler/knativeserving/ingress/kourier.go#L82-L84

I ended up refactoring the whole function. Let me know if you prefer this or if I should revert.

theomessin:
@nak3 any update on this?

config/crd/bases/operator.knative.dev_knativeservings.yaml
nak3: I am prefer to `service-load-balancer-ip`. WDYT @houshengbo ?

nak3 (commented):
Thank you! LGTM except for one comment.

config/crd/bases/operator.knative.dev_knativeservings.yaml
theomessin: Done

nak3:
/lgtm
/approve

/assign @houshengbo 
Thank you! I think I don't have a permission to merge this so defer to @houshengbo

nak3:
bump @houshengbo
/cc @dprotaso @psschwei

theomessin:
Are we only waiting on the lint failing? The error message doesn't mean anything to me...

```
Running [/home/runner/golangci-lint-1.52.0-linux-amd64/golangci-lint run --out-format=github-actions --new-from-patch=/tmp/tmp-1720-r08HIklCaYvx/pull.patch --new=false --new-from-rev= --go=1.20] in [] ...
  Error: unnecessary conversion (unconvert)
  ```

**Comment by theomessin** [comment]
@nak3 @houshengbo

File: pkg/reconciler/knativeserving/ingress/kourier_test.go [hunk 3]
```diff
@@ -158,6 +180,16 @@ func verifyGatewayServiceType(t *testing.T, u *unstructured.Unstructured, expSer
 	}
 }
 
+func verifyGatewayServiceLoadBalancerIP(t *testing.T, u *unstructured.Unstructured, expServiceLoadBalancerIP string) {
+	if u.GetKind() == "Service" && u.GetName() == kourierGatewayServiceName {
+		svc := &v1.Service{}
+		err := scheme.Scheme.Convert(u, svc, nil)
+		util.AssertEqual(t, err, nil)
+		svcLoadBalancerIP := svc.Spec.LoadBalancerIP
+		util.AssertDeepEqual(t, string(svcLoadBalancerIP), expServiceLoadBalancerIP)
```
nak3, line 189:
```suggestion
		util.AssertDeepEqual(t, svcLoadBalancerIP, expServiceLoadBalancerIP)
```

Could you fix the lint error?

nak3:
As I commented https://github.com/knative/operator/pull/1343#issuecomment-1588287110, I don't have a power to approve PR.

nak3:
/lgtm
/approve

(again I think I don't have a power to merge..)

psschwei (commented):
I don't know the operator but trust @nak3 's judgment 
/approve
/hold
In case @houshengbo wants to weigh in, but feel free to unhold in a day or two

theomessin:
@nak3 @psschwei @houshengbo Can we please get this merged? The longer it sits the more work I have to keep doing to resolve conflicts every time main changes...

PR #1355, opened 2023-02-27
Change file structure for ingresses

### LINKED ISSUE #1345: Change the directory stucture for ingresses
**Problem**
All the ingresses are saved in one directory. We need to decouple them into separate directories to make sure each ingress has a valid path to retrieve all its manifests. This leads to better performance, because it will read all the manifests for one certain ingress at one time. So far, we need to read all the manifests for all ingresses, and use a filter to remove the manifests, that are not in need.

Developer.

**Exit Criteria**
The proposed new structure is similar to the directory, saving the eventing sources.

**Time Estimate (optional):**
One week for one person.

None

houshengbo:
Fixes #1345

## Proposed Changes

* This PR aimed to change the selection mode for the ingresses from a filter based mode to a directory based mode. We can pick up the ingress by the name of the ingress from the designated directory

**Release Note**

```release-note

```


pkg/reconciler/knativeserving/ingress/kourier.go
```diff
@@ -140,3 +138,10 @@ func configureBootstrapConfigMap(instance *v1beta1.KnativeServing) mf.Transforme
 		return nil
 	}
 }
+
+func hasProviderLabel(u *unstructured.Unstructured) bool {
```
nak3, line 142: We can drop `hasProviderLabel` in Kourier as well so please feel free to drop it. Or I will drop them in the separate PR.

pkg/reconciler/knativeserving/ingress/ingress.go
```diff
@@ -87,61 +49,90 @@ func Transformers(ctx context.Context, ks *v1beta1.KnativeServing) []mf.Transfor
...
+		url := filepath.Join(ingressPath, "kourier")
+		urls = append(urls, url)
 	}
 
-	// This line can make sure a valid available ingress version is returned.
-	ingressVersion = common.GetLatestIngressRelease(ingressVersion)
-	ingressPath := filepath.Join(koDataDir, "ingress", ingressVersion)
-	return common.FetchManifest(ingressPath)
+	if len(urls) == 0 {
+		url := filepath.Join(ingressPath, "istio")
```
nak3, line 89: We have the default behavior around L#67 so it should be fine, I think. But currently (without this PR) `nonFilter` was returned and no ingress was created. So, it should be same behavior instead of installing Istio?

pkg/reconciler/knativeserving/ingress/ingress.go
```diff
@@ -87,61 +49,90 @@ func Transformers(ctx context.Context, ks *v1beta1.KnativeServing) []mf.Transfor
...
+	// If the user specified a version with a minor version, which is not supported by the current operator, the operator
+	// can still work, as long as spec.manifests contains all the manifest links. This function can always return nil,
 	// even if the ingress is not available.
 	return nil
 }
 
-func hasProviderLabel(u *unstructured.Unstructured) bool {
-	if _, hasLabel := u.GetLabels()[providerLabel]; hasLabel {
-		return true
+func convertToKS(instance base.KComponent) *v1beta1.KnativeServing {
```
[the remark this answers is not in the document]
houshengbo: good point.

pkg/reconciler/knativeserving/ingress/ingress.go
houshengbo: Right. Let me change it. I can add some more test cases to see if the edge cases are covered.

nak3:
/cc @skonto @ReToCode 

LGTM

cmd/fetcher/kodata/config.yaml
```diff
@@ -23,16 +23,19 @@ ingress:
   - s3:
       bucket: "gs-noauth://knative-releases"
       prefix: "net-istio/previous"
+    eventingService: istio
```
ReToCode, line 26: Is there a reason why this is called `eventingService`? I was first confused with Eventing as we are in the `knative-serving` part of the yaml.

ReToCode (commented):
Other than that, LGTM.

cmd/fetcher/kodata/config.yaml
skonto: IngressService or just ingress?

PR #1381, opened 2023-03-15
Allow to change webhook secret name via WEBHOOK_SECRET_NAME

hhk7734:
It is next action in #1377


houshengbo (commented):
/lgtm
/approve

houshengbo:
@hhk7734 Please join the list of contributors, so that we do not need the ok-to-test label any more.
This is an example: https://github.com/knative/community/pull/1264/files

PR #1447, opened 2023-05-18
Add the helm template and the instruction on how to release helm chart

### LINKED ISSUE #914: Add helm chart to release
**Problem**
Currently there is an [operator.yaml](https://github.com/knative/operator/releases/download/knative-v1.1.0/operator.yaml) attached to each release which is generated with bash scripts upon each release. However all namespaced resources in that release are installed in the `default` namespace which might not be suitable for larger production clusters.

We would like to install knative via a helm chart where the namespace is set to `{{ .Release.Namespace }}`.

The reason why I think this should be added upstream is because even if you make your own helm chart using the operator yaml from the release via `curl -sLo knative/templates/operator.yaml https://github.com/knative/operator/releases/download/knative-v1.1.0/operator.yaml`, then you still need to manually replace the namespaces and move the CRDs into a `crd` directory (otherwise helm can't install a `KnativeServing` resource in the same release.

System Operator

**Exit Criteria**
Helm chart is attached to the release and operators can run something like 
```
helm repo add knative https://knative.dev
helm install knative knative/knative
```

**Time Estimate (optional):**
2 days

houshengbo:
Fixes #914


houshengbo:
@jonnylangefeld could you review this pr? thx.

nabero:
Deployed on k3s and I confirm it did work fine at my end.

If I may suggest:
- As a user working in a confined environment, pulling via a proxified gcr.io would have been easier there was one parameter dedicated update the repository for all images.  As an example, istiod charts have a parameter "hub: docker.io/istio" which is used for all images and that can be overridden on each sub component separately.
- I suspect it should require changes on the operator itself, but I would have expected to be able to override that "hub" parameter at the operator level, and that subsequent deployment of KnativeEventing and KnativeServing would take it into account. 

Thanks !

PR #1452, opened 2023-05-23
Simplify the transformer for the installed resources

houshengbo:
Fixes: #1462

### DIFF
--- a/pkg/reconciler/common/transformers.go
+++ b/pkg/reconciler/common/transformers.go
@@ -57,8 +57,8 @@ func Transform(ctx context.Context, manifest *mf.Manifest, instance base.KCompon
 	return nil
 }
 
-// InstalledTransform will mutate the namespace of all installed resources
-func InstalledTransform(manifest *mf.Manifest, instance base.KComponent, extra ...mf.Transformer) error {
+// InjectNamespace will mutate the namespace of all installed resources
+func InjectNamespace(manifest *mf.Manifest, instance base.KComponent, extra ...mf.Transformer) error {
 	transformers := []mf.Transformer{
 		mf.InjectNamespace(instance.GetNamespace()),
 	}
--- a/pkg/reconciler/common/transformers.go
+++ b/pkg/reconciler/common/transformers.go
@@ -56,3 +56,18 @@ func Transform(ctx context.Context, manifest *mf.Manifest, instance base.KCompon
 	*manifest = m
 	return nil
 }
+
+// InstalledTransform will mutate the namespace of all installed resources
+func InstalledTransform(manifest *mf.Manifest, instance base.KComponent, extra ...mf.Transformer) error {
+	transformers := []mf.Transformer{
+		mf.InjectNamespace(instance.GetNamespace()),
+	}
+	transformers = append(transformers, extra...)
+	m, err := manifest.Transform(transformers...)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


matzew:
I miss a bit of description for this PR. Also why is it done. Is there some GH issue that describes the problem?
Or at least a bit more in the PR?

houshengbo:
I opened an issue, describing why the this PR is needed.

pkg/reconciler/common/transformers.go
```diff
@@ -56,3 +56,18 @@ func Transform(ctx context.Context, manifest *mf.Manifest, instance base.KCompon
...
+
+// InstalledTransform will mutate the namespace of all installed resources
+func InstalledTransform(manifest *mf.Manifest, instance base.KComponent, extra ...mf.Transformer) error {
+	transformers := []mf.Transformer{
+		mf.InjectNamespace(instance.GetNamespace()),
+	}
+	transformers = append(transformers, extra...)
+	m, err := manifest.Transform(transformers...)
+	if err != nil {
+		instance.GetStatus().MarkInstallFailed(err.Error())
```
nak3, line 68: `MarkInstallFailed` is a little bit strange for me. It would be better to have another status which is ignorable. It should be another PR though.

pkg/reconciler/common/transformers.go
```diff
@@ -56,3 +56,18 @@ func Transform(ctx context.Context, manifest *mf.Manifest, instance base.KCompon
 	*manifest = m
 	return nil
 }
+
+// InstalledTransform will mutate the namespace of all installed resources
+func InstalledTransform(manifest *mf.Manifest, instance base.KComponent, extra ...mf.Transformer) error {
```
nak3, line 61: Since this is just injecting the namespace to `manifest`, the function name `InstalledTransform` should be `injectNamespace` or something?

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -146,14 +146,21 @@ func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, comp
...
+
 func (r *Reconciler) installed(ctx context.Context, instance base.KComponent) (*mf.Manifest, error) {
 	paths := instance.GetStatus().GetManifests()
 	installed, err := common.FetchManifestFromArray(paths)
 	if err != nil {
 		return &installed, err
 	}
 	installed = r.manifest.Append(installed)
-	stages := common.Stages{r.transform}
+	stages := common.Stages{r.installedTransform}
```
nak3, line 163: Could you add a comment that the staged manifest is just added the `Namespace` because of https://github.com/knative/operator/issues/1462 ?

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -146,14 +146,21 @@ func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, comp
 	return common.Transform(ctx, manifest, instance, extra...)
 }
 
+// installedTransform mutates the namespace of all installed resources
+func (r *Reconciler) installedTransform(ctx context.Context, manifest *mf.Manifest, comp base.KComponent) error {
```
nak3, line 150: Same with above, `installedTransform` just mutates the namespace so the function name should just be more specific.

PR #1479, opened 2023-06-22
Enable the support of NodeSelectors and tolerations on Jobs

houshengbo:
Fixes: #1480

### DIFF
--- a/pkg/reconciler/common/workload_override.go
+++ b/pkg/reconciler/common/workload_override.go
@@ -20,6 +20,7 @@ import (
 	mf "github.com/manifestival/manifestival"
 	"go.uber.org/zap"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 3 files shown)


pkg/reconciler/common/workload_override.go
```diff
@@ -61,6 +62,14 @@ func OverridesTransform(overrides []base.WorkloadOverride, log *zap.SugaredLogge
 					ss.Spec.Replicas = override.Replicas
 				}
 			}
+			if u.GetKind() == "Job" && u.GetGenerateName() == override.Name {
+				ss := &batchv1.Job{}
```
nak3, line 66:
```suggestion
				job := &batchv1.Job{}
```
nit: `ss` is used for `StatefulSet` above but here is job.

nak3 (commented):
LGTM except for one nit.

PR #1503, opened 2023-07-26
Use kubeVersion to constrain kubernetes version when using helm

hhk7734:
## Proposed Changes

I moved the Kubernetes version restriction from values.yaml to Chart.yaml because I thought it would be nice to be limited by Helm.


houshengbo:
@hhk7734 The reason I put it in values is to make it configurable by different platforms. Like in my clusters, I need to override this value.

hhk7734:
@houshengbo Is it better to leave both `kubeVersion` and `kubernetes_min_version`? Or should I close this PR?

As an additional question, is there a reason for using `snake case`? Most of the helm charts I've seen so far are in `camel case`, so it feels out of place. Also, I'm wondering if there's a reason why the `apiVersion` is `v1`. I know v1 is recognized by helm3, but I don't see the need to support older versions.

https://helm.sh/docs/topics/version_skew/

PR #1519, opened 2023-08-03
Allow to disable probe

nak3:
## Proposed Changes

Currently it is not possible to disable {readiness,liveness}Probes. It is a problem for kourier gateway.
(Kourier's proble uses admin interface so users cannot change the admin config (`admin.address`) for debug without disabling the probe. More specifically please find `/tmp/envoy.admin` in [bootstrap.yaml](https://github.com/knative-extensions/net-kourier/blob/main/config/200-bootstrap.yaml) and [300-gateway.yaml](https://github.com/knative-extensions/net-kourier/blob/main/config/300-gateway.yaml))

Hence this patch changes to disable probe when users explicitly set the empty overrideProbe.

For example, if they have the following setting:

```yaml
  workloads:
  - livenessProbes:
    - container: kourier-gateway
    name: 3scale-kourier-gateway
```

then, drop livenessProbes from 3scale-kourier-gateway deployment.

**Release Note**

```release-note
Disable probe when explicitly settting the empty overrideProbe.
```

### DIFF
--- a/pkg/reconciler/common/workload_override.go
+++ b/pkg/reconciler/common/workload_override.go
@@ -174,7 +174,8 @@ func replaceProbes(override *base.WorkloadOverride, ps *corev1.PodTemplateSpec) 
 	if len(override.ReadinessProbes) > 0 {
 		containers := ps.Spec.Containers
 		for i := range containers {
-			if override := findProbeOverride(override.ReadinessProbes, containers[i].Name); override != nil {
+			override := findProbeOverride(override.ReadinessProbes, containers[i].Name)
+			if override != nil {
 				overrideProbe := &v1.Probe{
 					InitialDelaySeconds:           override.InitialDelaySeconds,
 					TimeoutSeconds:                override.TimeoutSeconds,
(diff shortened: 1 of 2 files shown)


pkg/reconciler/common/workload_override.go
```diff
@@ -183,6 +184,11 @@ func replaceProbes(override *base.WorkloadOverride, ps *corev1.PodTemplateSpec)
 					FailureThreshold:              override.FailureThreshold,
 					TerminationGracePeriodSeconds: override.TerminationGracePeriodSeconds,
 				}
+				if *overrideProbe == (v1.Probe{}) {
```
houshengbo, line 187: I am not sure if my logic got something wrong:
Line 179 overrideProbe := &v1.Probe{...}, *overrideProbe is for sure equal to (v1.Probe{}). so Line 192 will never reach.

pkg/reconciler/common/workload_override.go
```diff
@@ -204,6 +210,11 @@ func replaceProbes(override *base.WorkloadOverride, ps *corev1.PodTemplateSpec)
...
```
houshengbo, line 213: The same to the above comment.

pkg/reconciler/common/workload_override.go
nak3: L192 will still reachable.

- The new logic happens only when users didn't specify any override probe. (=`overrideProbe` is empty.)
- L192 will be reached when users specify some override probe && the target manifest does not have any probe. (=`overrideProbe` is not empty but the manifest's `containers[i].ReadinessProbe` is empty.)

PR #1534, opened 2023-08-17
Add more env vars for pingsource adapter env var preservation

aliok:
Fixes #

## Proposed Changes

* Add more env vars for pingsource adapter env var preservation
* Without this, the eventing-controller and the operator are fighting to set env var values
* Related to https://github.com/knative/operator/issues/393
* Ideally, we should find a way to fix this problem for good! Current solution is to maintain the env vars manually.

**Release Note**

```release-note

```


matzew (approved):
/lgtm
/approve

matzew:
@aliok we might need backports?

PR #1541, opened 2023-08-17
Added Ports to NodePort service-types in knative-serving

### LINKED ISSUE #1523: The external Ports in kourier should be configurable. (change port/disable external ports)
**Problem**
In my cluster the ports 80 and 443 are already taken by another service (traefik). As such the kourier LoadBalancer request for assigning the ports are pending. It would be great that the ports for kourier, which are configured in the Service could be changed or, that the external ports could be disabled (then I would configure traefik to forward to the internal service of kourier)

[Persona:Operator]

**Exit Criteria**
The external kourier ports could be configured (disabled or set to other values) by the KnativeServing type. The ports will be bound to the configured ports or aren't bound to any external interface (in case of disabling)

eBeyond:
/lint

Fixes #1523

## Proposed Changes

For the ServiceType NodePort it's now possible to add http-port and https-port. So the NodePorts of the service can be specified in the case that the standard isn't suitable.
**Release Note**

```release-note
Added http-port and https-port for the ServiceType NodePort  (kourier config)

```


houshengbo:
@eBeyond Plz fix all Golint naming issues.
You need to change the CRD as well, in order to add the fields.

eBeyond:
Changed the Http/Https to HTTP/HTTPS as requested by the linter. 
Added the fields to the CRD

PR #1548, opened 2023-08-30
set HPA replica values when workload overrides are set

### LINKED ISSUE #762: `spec.deployments.replicas` does not take care about HorizontalPodAutoscaler

ReToCode:
Fixes #762

## Proposed Changes
* Set HPA `minReplicas` and `maxReplicas` when `workloads.replicas` overrides are present for an object with HPA
* Uses the same behaviour as for the `high-availability` field
* Moved HPA transformation logic to a common file
* Add Eventing HPAs to `hasHorizontalPodAutoscaler`
* Add HPA name mapping for Eventing HPAs that are named different to their target: `getHPAName`

**Release Note**
```release-note
The operator now sets HorizontalPodAutoscaler replicas (on resources with HPAs) when workload overrides are defined.
```

### DIFF
--- a/pkg/reconciler/common/ha_test.go
+++ b/pkg/reconciler/common/ha_test.go
@@ -23,7 +23,6 @@ import (
 	"knative.dev/operator/pkg/apis/operator/v1beta1"
 
 	appsv1 "k8s.io/api/apps/v1"
-	autoscalingv2beta1 "k8s.io/api/autoscaling/v2beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -74,7 +73,7 @@ func TestHighAvailabilityTransform(t *testing.T) {
 		in:       makeUnstructuredHPA(t, "activator", 2, 5),
 		expected: makeUnstructuredHPA(t, "activator", 2, 5),
 	}, {
-		name:     "HA; adjust hpa when replicas is lerger than maxReplicas",
+		name:     "HA; adjust hpa when replicas is larger than maxReplicas",
 		config:   makeHa(6),
 		in:       makeUnstructuredHPA(t, "activator", 2, 5),
 		expected: makeUnstructuredHPA(t, "activator", 6, 9), // maxReplicas is increased by max+(replicas-min) to avoid minReplicas > maxReplicas happenning.
@@ -107,7 +106,7 @@ func TestHighAvailabilityTransform(t *testing.T) {
 					},
 				},
 			}
-			haTransform := HighAvailabilityTransform(instance, log)
+			haTransform := HighAvailabilityTransform(instance)
 			err := haTransform(tc.in)
 
 			util.AssertDeepEqual(t, err, tc.err)
@@ -139,26 +138,6 @@ func makeUnstructuredDeploymentReplicas(t *testing.T, name string, replicas int3
 	err := scheme.Scheme.Convert(d, result, nil)
 	if err != nil {
 		t.Fatalf("Could not create unstructured Deployment: %v, err: %v", d, err)
-	}
-
-	return result
-}
-
-func makeUnstructuredHPA(t *testing.T, name string, minReplicas, maxReplicas int32) *unstructured.Unstructured {
-	hpa := &autoscalingv2beta1.HorizontalPodAutoscaler{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
-		},
-		Spec: autoscalingv2beta1.HorizontalPodAutoscalerSpec{
-			MinReplicas: &minReplicas,
-			MaxReplicas: maxReplicas,
-		},
-	}
-
-	result := &unstructured.Unstructured{}
-	err := scheme.Scheme.Convert(hpa, result, nil)
-	if err != nil {
-		t.Fatalf("Could not create unstructured HPA: %v, err: %v", hpa, err)
 	}
 
 	return result
--- a/pkg/reconciler/common/ha_test.go
+++ b/pkg/reconciler/common/ha_test.go
@@ -76,7 +76,7 @@ func TestHighAvailabilityTransform(t *testing.T) {
 		name:     "HA; adjust hpa when replicas is larger than maxReplicas",
 		config:   makeHa(6),
 		in:       makeUnstructuredHPA(t, "activator", 2, 5),
-		expected: makeUnstructuredHPA(t, "activator", 6, 9), // maxReplicas is increased by max+(replicas-min) to avoid minReplicas > maxReplicas happenning.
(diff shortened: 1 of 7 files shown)


pkg/reconciler/common/workload_override.go
```diff
@@ -71,6 +75,13 @@ func OverridesTransform(overrides []base.WorkloadOverride, log *zap.SugaredLogge
 				ps = &job.Spec.Template
 			}
 
+			if u.GetKind() == "HorizontalPodAutoscaler" && u.GetName() == override.Name && override.Replicas != nil {
```
nak3, line 78: Current `[]base.WorkloadOverride` (`spec.workloads`) assume that it contains only `Deployment`, `StatefulSet`, and `Job`, so this condition does not happen. 

We **may** able to support PDB for `spec.workloads` but it is weird that users need to set `override.Replicas` for PDB.

nak3: Ah, I guess the code assume `Deployment`'s name is same with `PDB`'s name? If so, it is different like https://github.com/knative/serving/blob/ba5a9298901945907498c7e6b9c28a09bc346f59/config/core/deployments/webhook-hpa.yaml#L44

nak3: Ah, the code assumes that `Deployment` name is same with `HorizontalPodAutoscaler`'s name?

ReToCode: Yes, AFAIK this is the current assumption of the operator. This PR just introduces the same same behaviour as for the `high-availability` flag. So the example for `mt-channel-broker` would also currently not work with `high-availability` set.

We probably should introduce a new concept to configure the HPAs with min+max fields anyways, but that is another story (mid to long term). So for the short term, can eventing change the name of the HPA? @matzew , @pierDipi

pierDipi: Changing the name requires some form of migration, can't we have a mapping `deployment` -> `hpa name` and use that?

pkg/reconciler/common/ha_test.go
```diff
@@ -74,7 +73,7 @@ func TestHighAvailabilityTransform(t *testing.T) {
 		in:       makeUnstructuredHPA(t, "activator", 2, 5),
 		expected: makeUnstructuredHPA(t, "activator", 2, 5),
 	}, {
-		name:     "HA; adjust hpa when replicas is lerger than maxReplicas",
+		name:     "HA; adjust hpa when replicas is larger than maxReplicas",
 		config:   makeHa(6),
 		in:       makeUnstructuredHPA(t, "activator", 2, 5),
 		expected: makeUnstructuredHPA(t, "activator", 6, 9), // maxReplicas is increased by max+(replicas-min) to avoid minReplicas > maxReplicas happenning.
```
skonto, line 79: typo: happenning

pkg/reconciler/common/hpa.go
```diff
@@ -0,0 +1,71 @@
...
+		return err
+	}
+
+	// Do nothing if maxReplicas is not defined.
+	if !found {
+		return nil
+	}
+
+	// Increase maxReplicas to the amount that we increased,
+	// because we need to avoid minReplicas > maxReplicas happenning.
```
skonto, line 66: typo: happenning

pkg/reconciler/common/hpa.go
```diff
@@ -0,0 +1,71 @@
...
+	if err := unstructured.SetNestedField(u.Object, max+(replicas-min), "spec", "maxReplicas"); err != nil {
```
skonto, line 67: It would be nice if we log something here for debugging reasons.

pkg/reconciler/common/workload_override.go
skonto: Let's do a follow up PR to fix the inconsistencies with the names.

nak3: > Yes, AFAIK this is the current assumption of the operator. This PR just introduces the same same behaviour as for the high-availability flag. 

I think `high-availability` (=`HighAvailabilityTransform()`) does not have the assumption. It applies min/max replica settings to all HPA regardless of the name.
(Note `hasHorizontalPodAutoscaler()` is to avoid another issue, not for the inconsistent name.)

ReToCode: I agree with you from the code, but it still has a bit the assumption, as `hasHorizontalPodAutoscaler` would currently interfere with the eventing HPAs, so I added the mapping @pierDipi proposed. Now all three cases should be fine:

**Given** 
- mt-broker-ingress: replicas are not set in yaml
- broker-ingress-hpa: 1-10

**Case 1** 
- high-available=1
- no workload overrides

HighAvailabilityTransform
- mt-broker-ingress —> skipped
- broker-ingress-hpa —> not skipped, _set to 1_

OverridesTransform
- mt-broker-ingress —> skipped
- broker-ingress-hpa —> skipped

**Case 2**
- high-available not set
- workload overrides: mt-broker-ingress=2

HighAvailabilityTransform
- mt-broker-ingress —> skipped
- broker-ingress-hpa —> skipped

OverridesTransform
- mt-broker-ingress —> _set to 2_
- broker-ingress-hpa —> _set to 2_


**Case 3**
- high-available=3
- workload overrides: mt-broker-ingress=2

HighAvailabilityTransform
- mt-broker-ingress —> skipped
- broker-ingress-hpa —> not skipped, _set to 3_

OverridesTransform (this is executed later, as it is more specific)
- mt-broker-ingress —> _set to 2_
- broker-ingress-hpa —> _set to 2_

skonto: Are these all covered in tests?

ReToCode: I think so, but I'll double-check.

pierDipi: Thanks, followup is fine as well

skonto:
LGTM @houshengbo should we merge?
@ReToCode does this need any doc updates here https://knative.dev/docs/install/operator/configuring-serving-cr/#high-availability or we are good?

ReToCode:
> @ReToCode does this need any doc updates here https://knative.dev/docs/install/operator/configuring-serving-cr/#high-availability or we are good?

No that should be fine. This does not change anything for the `high-available` field.

pkg/reconciler/common/workload_override.go
ReToCode: > Are these all covered in tests?

Now all those cases are covered in tests.

ReToCode:
@houshengbo would you mind taking a look?

PR #1571, opened 2023-09-27
chore: Add resources requests and limits

kahirokunn:
Fixes #

## Proposed Changes

* If resources requests are left out, they will be given a lower priority when scheduled, so we have assigned them to all.
*
*

**Release Note**

```release-note
chore: Add resources requests and limits
```

### DIFF
--- a/config/charts/knative-operator/templates/operator.yaml
+++ b/config/charts/knative-operator/templates/operator.yaml
@@ -6245,5 +6245,12 @@ spec:
           ports:
             - name: metrics
               containerPort: 9090
+          resources:
+            requests:
+              memory: 32Mi
+              cpu: 20m
+            limits:
+              memory: 100Mi
+              cpu: 50m
 
 ---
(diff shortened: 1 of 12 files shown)


config/charts/knative-operator/templates/operator.yaml
```diff
@@ -6245,5 +6245,12 @@ spec:
           ports:
             - name: metrics
               containerPort: 9090
+          resources:
```
matzew, line 6248: Do these numbers come from any verification or such?

kahirokunn: Yes, I have verified this in my environment on a metrics basis and it works well with this value.

kahirokunn: If it's not a problem, I'll try to fix the unit test.

kahirokunn: @matzew Your opinion.
Does it seem safe to proceed with implementation?
Thx.

houshengbo:
@kahirokunn we ultimately added `requests and limits`, as in https://github.com/knative/operator/blob/main/config/manager/operator.yaml#L41

kahirokunn:
Thanks!

PR #1589, opened 2023-10-09
Improve manifest deletion error handling

pierDipi:
It might happen that not all resources were created successfully so we should not fail finalization when resources are not found, we can just ignore not found errors.


creydr (approved):
/lgtm

pierDipi:
Cherry pick done manually in https://github.com/knative/operator/pull/1591

PR #1590, opened 2023-10-09
Improve KnativeEventing finalization for TLS resources

pierDipi:
For optional resources like cert-manager's Certificates and Issuers, we don't
want to fail finalization when such operators are not installed, so during
finalization, we split the resources in
- optional resources (TLS resources, etc)
- resources (core k8s resources)

Then, we delete `resources` first and after we delete optional resources
while also ignoring errors returned when such operators are not instaled.


creydr (approved):
/lgtm

pierDipi:
Done manually https://github.com/knative/operator/pull/1591, bot is broken https://github.com/kubernetes/test-infra/issues/30937

PR #1604, opened 2023-10-19
Move to Security-Guard 0.6

davidhadas:
Moved to Security-Guard 0.6

PR #1611, opened 2023-10-24
Move to Security-Guard 0.6 

houshengbo:
Bump Security Guard to 0.6


houshengbo:
@davidhadas This is probably need you to change.

davidhadas:
@houshengbo , 
I am not sure what you did to reach here, but this includes all changes in #1604 and therefore I am closing #1604 and replacing it with this one. 

Thanks.
Are we expected to face a similar issue when we will bump to 0.7? If so, can you describe what was needed to get here?
I.e. can we document here or elsewhere the procedure to bump Security Advisor version in operator?

PR #1638, opened 2023-11-29
manual bump, by running './hack/update-deps.sh --upgrade'

matzew:
Fixes #

## Proposed Changes

* manual bump to latest. Done instead of #1633 which is also bumping the golang version... 

**Release Note**

```release-note

```


ReToCode:
@matzew can you rerun `hack/update-codegen.sh`?

PR #1674, opened 2024-01-17
Eventing TLS: Add trust manager resources to filter

pierDipi:
Eventing TLS will have trust manager resources in the artifacts and we need to manage them as the core cert-manager resources.


creydr (approved):
/lgtm

creydr (approved):
/lgtm

PR #1712, opened 2024-02-21
Fix the cluster version to 1.28

houshengbo:


config/manager/operator.yaml
```diff
@@ -55,7 +55,7 @@ spec:
             - name: CONFIG_OBSERVABILITY_NAME
               value: config-observability
             - name: KUBERNETES_MIN_VERSION
-              value: ""
+              value: "1.24.0"
```
matzew: why was it empty before?

houshengbo: It was and will be empty. We would like to offer a way to override the kube version. If it is empty, it will default to the value knative/pkg sets.

PR #1732, opened 2024-03-12
Removed eventing-tls-networking.yaml from the manifests

houshengbo:
## Proposed Changes

* KnativeEventing does not install cert-manager by default. We have to remove eventing-tls-networking.yaml from the core.


pierDipi:
/hold

why?

houshengbo:
@pierDipi By default, the feature flag seems to be on, so the reconcile loop of operator runs into 
```
ClusterIssuer” in version “[cert-manager.io/v1](http://cert-manager.io/v1)`
```
However, if folks do not know cert-manager is the dependency, they would think it is a bug on operator.

Per the discussion here: https://cloud-native.slack.com/archives/C04LMU33V1S/p1710250853335799, we need a better support of this TLS feature in later release.
But for now, what is your suggestion to fix?

pierDipi:
That's the problem, it's shouldn't be on by default

PR #1767, opened 2024-04-11
Perf: add resource limits to operators.

kycheng:
I think increasing the resource limit is a better best practice to prevent pods from using resources infinitely (this is not currently happening.)

Based on the monitoring data of our environment, I set the request and limit values to 100m/100Mi

image

## Proposed Changes

* knative-operator should set resource limits

**Release Note**

```release-note
NONE
```

### DIFF
--- a/config/charts/knative-operator/templates/operator.yaml
+++ b/config/charts/knative-operator/templates/operator.yaml
@@ -6242,6 +6242,13 @@ spec:
               value: config-observability
             - name: KUBERNETES_MIN_VERSION
               value: "{{ .Values.knative_operator.kubernetes_min_version }}"
+          resources:
+            requests:
+              cpu: 100m
+              memory: 100Mi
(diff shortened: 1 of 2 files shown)


houshengbo:
What is the reason for adding resource limit? Why is it just 100?

kycheng:
> What is the reason for adding resource limit? Why is it just 100?

Sorry I forgot to add the reason.

I think increasing the resource limit is a better best practice to prevent pods from using resources infinitely (this is not currently happening.)

100 From the monitoring data of our operating environment, it seems that 100 is enough.

image

ReToCode:
> 100 From the monitoring data of our operating environment, it seems that 100 is enough.

This might be true for your environment, but not for everyones. I'd argue that this is a concern each operator has to take care of and figure out the correct values for their installation.

PR #1773, opened 2024-04-23
Change the namespace into knative-operator

### LINKED ISSUE #1725: Operator install into "default" namespace
https://knative.dev/docs/install/operator/knative-with-operators/#verifying-image-signatures 
By default operator install into namespace default. 
kubectl apply -f https://github.com/knative/operator/releases/download/knative-v1.13.2/operator.yaml

**Expected behavior**
It will be expected that no namespace will be hardcoded so user can use manifests or another namespace is hardcoded f.e. "knative-operator".  
Screenshot 2024-03-06 at 10 19 07
IMHO, hardcode "knative-operator" as a default namespace is looks better and user can redefine it with kustomize if needed. 
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
metadata:
  name: knative-operator

resources:
- operator.yaml

namespace: knative-operator-v2
```
**Additional context**
If there are no specific reason of deploying it into default namespace, I can change it.

houshengbo:
Fixes #1725

### DIFF
--- a/config/crd/bases/operator.knative.dev_knativeeventings.yaml
+++ b/config/crd/bases/operator.knative.dev_knativeeventings.yaml
@@ -2367,5 +2367,5 @@ spec:
       clientConfig:
         service:
           name: operator-webhook
-          namespace: default
+          namespace: knative-operator
           path: /resource-conversion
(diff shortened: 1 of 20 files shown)


kahirokunn:
That's awesome! LGTM

Issue #1869, opened 2024-08-19
Make securitycontext and resources configurable, annotations, lables configurable as well in Helm


houshengbo:
Helm should offer more options for the users to configure securitycontext and resources, annotations, and lables.

houshengbo:
This issue is fixed by https://github.com/knative/operator/pull/1911.

PR #1872, opened 2024-08-22
[release-1.14]: Configure the securityContext, resources and pod annotations in helm

houshengbo:
Fixed: #1869

### DIFF
--- a/config/charts/knative-operator/templates/operator.yaml
+++ b/config/charts/knative-operator/templates/operator.yaml
@@ -83,6 +83,10 @@ spec:
               topologyKey: kubernetes.io/hostname
             weight: 100
 
+      {{- if and .Values.knative_operator.operator_webhook.securityContext }}
+      securityContext:
+        {{ toYaml .Values.knative_operator.operator_webhook.securityContext }}
+      {{- end }}
       serviceAccountName: operator-webhook
       containers:
       - name: operator-webhook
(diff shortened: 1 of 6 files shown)


config/charts/knative-operator/templates/operator.yaml
```diff
@@ -48,6 +48,10 @@ metadata:
     app.kubernetes.io/component: operator-webhook
     app.kubernetes.io/version: "{{ .Chart.Version }}"
     app.kubernetes.io/name: knative-operator
+  {{- if and .Values.knative_operator.operator_webhook.annotations }}
```
yuzisun, line 51: I think the annotations under the metadata is not necessarily the same as the annotations on the template/metadata.

houshengbo: Just remove this section.

config/charts/knative-operator/templates/operator.yaml
```diff
@@ -114,14 +130,23 @@ spec:
           value: knative.dev/operator
         - name: KUBERNETES_MIN_VERSION
           value: "{{ .Values.knative_operator.kubernetes_min_version }}"
+        {{- if and .Values.knative_operator.operator_webhook.securityContext }}
+        securityContext:
```
yuzisun, line 130: There is also another securityContext on the pod level

houshengbo: but the yaml does not have it be default, if .Values.knative_operator.operator_webhook.securityContext is not available, do not add it.

yuzisun (approved):

PR #1482, opened 2023-06-24
Feat: Istio gateway TLS support added

supu2:
Fixes #1428
## Proposed Changes

* The change allow manual TLS setup for Istio gateway.
* The changes tested on Knative operator 1.9.5 

## Example Configuration
```
apiVersion: operator.knative.dev/v1beta1
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  ingress:
    istio:
      enabled: true
      knative-ingress-gateway:
        selector:
          istio: ingressgateway
        servers:
        - port:
            number: 80
            name: http
            protocol: HTTP
            target_port: 80
          hosts:
          - '*'
        - port:
            number: 443
            name: https
            protocol: HTTPS
            target_port: 8443
          tls:
            mode: SIMPLE
            credentialName: tls-certificate
          hosts:
          - '*'
  config:
    istio:
      local-gateway.knative-serving.knative-local-gateway: "knative-local-gateway.istio-system.svc.cluster.local"
```

PR #1888, opened 2024-09-19
Allow HTTP->HTTPS redirection on Istio Gateway

penguinush:
Extends https://github.com/knative/operator/pull/1482.

One of the ServerTLSSettings on Istio Gatway is `httpsRedirect` that, according to [docs](https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings), redirects HTTP traffic to HTTPS. Usually this is a desired behaviour for production setup, the only alternative would be to not create HTTP listener at all.

PR leverages combination of `nullable` property and `oneOf` with `required` fields - this works fine currently as there is 2 distinct options (`mode: <>`, which requires `credentialName` to be set, or `httpsRedirect`, which is mutually exclusive with every other option), but it might become cumbersome if more options of different combinations are added.

Tested with the following specs:
```redirect-on
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - '*.example.com'
    port:
      name: http
      number: 80
      protocol: HTTP
    tls:
      httpsRedirect: true
  - hosts:
    - '*.example.com'
    port:
      name: https-production
      number: 443
      protocol: https
    tls:
      credentialName: example-tls
      mode: SIMPLE
```
and
```redirect-off
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - '*.example.com'
    port:
      name: http
      number: 80
      protocol: HTTP
  - hosts:
    - '*.example.com'
    port:
      name: https-production
      number: 443
      protocol: https
    tls:
      credentialName: example-tls
      mode: SIMPLE
```
Both produce correct Gateways.

Kubernetes 1.30 (EKS), Knative Operator 1.15.4, Knative Serving 1.15.0, Istio 1.23.0.

## Proposed Changes

* Allows HTTP listener on Istio Gateway to be configured with automatic HTTP->HTTPS redirection.

**Release Note**

```release-note
feature: Istio Gateway can be configured with automatic HTTP->HTTPS redirection
```


houshengbo:
@penguinush could you rebase this pr?

PR #1895, opened 2024-09-24
Use istio.sidecar.inject label instead of annotation

ReToCode:
Partially https://github.com/knative/serving/issues/15521

## Proposed Changes
* Use istio.sidecar.inject label instead of annotation

/assign @skonto


matzew (approved):
/lgtm
/approve

Issue #1944, opened 2024-12-18
Make the HTTP to HTTPS redirection on the Istio Gateway configurable through the Helm chart


AyushSawant18588:
**Problem**
Currently through knative operator helm chart HTTP->HTTPS redirection on Istio Gateway is not configurable. This was done only in Knative operator manifest file in this PR: https://github.com/knative/operator/pull/1912. Can this be added in helm chart also so that http redirection can be configured in helm chart installation as well.

houshengbo:
@AyushSawant18588  Feel free to open a PR for this request.

AyushSawant18588:
> @AyushSawant18588 Feel free to open a PR for this request.

okay I will open a PR for this

houshengbo:
Fixed by https://github.com/knative/operator/pull/1948. Feel free to reopen it, if there is any issue.

PR #1948, opened 2024-12-23
Configure httpRedirect on istio gateway in helm

AyushSawant18588:
Fixes #
Update knative-serving CRD to allow configuring of HTTP->HTTPS redirection on Istio Gateway through helm chart. Currently through knative operator helm chart HTTP->HTTPS redirection on Istio Gateway is not configurable. This was done only in Knative operator manifest file in this PR: https://github.com/knative/operator/pull/1912

Issue: https://github.com/knative/operator/issues/1944

### DIFF
--- a/config/charts/knative-operator/templates/operator.yaml
+++ b/config/charts/knative-operator/templates/operator.yaml
@@ -4875,6 +4875,13 @@ spec:
                                       type: string
                                   type: object
                                 tls:
+                                  nullable: true
+                                  oneOf:
+                                  - required:
+                                    - mode
+                                    - credentialName
+                                  - required:
+                                    - httpsRedirect
                                   properties:
                                     mode:
                                       description: TLS mode can be SIMPLE, MUTUAL, ISTIO_MUTUAL.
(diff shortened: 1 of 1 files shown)


houshengbo:
@AyushSawant18588 could you join the knative org to pass the EasyCLA checking?

config/charts/knative-operator/templates/operator.yaml
```diff
@@ -4875,6 +4875,13 @@ spec:
                                       type: string
                                   type: object
                                 tls:
```
houshengbo, line 4877: I am wondering whether you also need to change `https://github.com/knative/operator/blob/main/config/charts/knative-operator/values.yaml` for the default values of these new fields.

AyushSawant18588: These are changes in Knative Serving CRDs so I do not think there will be any change in values.yaml of knative-operator helm chart. 
These changes will allow for httpRedirect configuration while creating Knative Serving Custom Resource after installing knative-operator through this helm chart. What do you think?

houshengbo: I thought you were about to configure the values for the fields added in this PR.
If they are not editable in values.yaml, they will always be the same values as defined in the CRD.

AyushSawant18588: This PR introduces changes to enable applying the httpRedirect setting when creating Knative Serving after installing the Knative Operator using the Helm chart. Please note that this setting is not to be configurable through the values.yaml file of the Knative Operator Helm chart since this setting is for Knative Serving.

PR #1982, opened 2025-01-31
feat(helm): Create .gitignore in gh-pages

patrickleet:
## Proposed Changes

* The helm publishing action expects this .gitignore file to exist - the script creates it on the first run, but the `gh-pages` branch was created manually and so that part never ran.

**Release Note**

```release-note
NONE
```


houshengbo (commented):
/lgtm
/approve

PR #2012, opened 2025-03-06
Fix Istio setup

skonto:
Fixes #

## Proposed Changes

* See #2011


skonto:
Verify should be fixed with #2011

skonto:
cc @houshengbo

PR #2013, opened 2025-03-06
Fix the Istio setup in CI

houshengbo:


matzew:
Is this different than the earlier PR: https://github.com/knative/operator/pull/2012 ? :sweat_smile:

PR #2016, opened 2025-03-11
Unset env.ValueFrom in ImageTransform

pierDipi:
When there is an environment variable override for ImageTransform, there could be `ValueFrom` set, in which case the resulting environment field would have both `Value` and `ValueFrom` which is not valid.


matzew (approved):
/lgtm
/approve

houshengbo:
@pierDipi I just fixed the CI issue, could you rebase this PR?

pierDipi:
done

PR #2044, opened 2025-03-22
fix: publish all helm charts

### LINKED ISSUE #2041: Helm chart: script only gets previous 30 releases
**Describe the bug**
In the Helm chart that's published each release, the bottom releases are removed when merged because results from the API are limited to 30 by default.

**Expected behavior**
All releases are indexed.

**To Reproduce**
See last release PR (https://github.com/knative/operator/pull/2033/files)

**Knative release version**
Each new version

**Additional context**
Should be an easy fix

As a first step, looks like we can increase the number of results returned by the API to 100. Beyond that, the script will need to handle pagination.

patrickleet:
The script previously only returned the last 30 releases, resulting in older releases being removed with each new release - it now uses pagination to get all of the releases.

Fixes #2041 

## Proposed Changes

* Adds pagination to script to get all releases

**Release Note**

```release-note
NONE
```

### DIFF
--- a/hack/find_helm_chart_releases_and_create_helm_index.sh
+++ b/hack/find_helm_chart_releases_and_create_helm_index.sh
@@ -15,22 +15,30 @@ entries:
   $CHART_NAME:
 EOF
 
-# Function to fetch all releases and filter .tgz files
 fetch_tgz_assets() {
   echo "Fetching release assets from GitHub API..." >&2
-  
-  # Get response from GitHub API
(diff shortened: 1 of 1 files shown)


patrickleet:
@houshengbo

patrickleet:
https://github.com/knative/operator/pull/2033/files#diff-08c65c1b5ef8384ef0200558777c52f4110afcbbacc880ac5375d555ca97ee3eL256-L273

fixes that ^

PR #2106, opened 2025-06-19
Propagate manifests paths earlier to operator status

dsimansk:
Fixes #

## Proposed Changes

* Propagate manifests paths earlier to operator status

```release-note
Propagate manifests paths earlier to operator status
```

A background, that could be capture in the issue, but pls bear with me. 

A common error pattern from status or operator logs:
```
    - lastTransitionTime: "2025-05-12T08:03:24Z"
      message: 'Install failed with message: Internal error occurred: failed calling
        webhook "config.webhook.serving.knative.dev": failed to call webhook: Post
        "https://webhook.knative-serving.svc:443/config-validation?timeout=10s": service
        "webhook" not found'
      reason: Error
      status: "False"
      type: InstallSucceeded 
```

In some cases of failed operator installation on OpenShift we have seen stuck install or reinstall attempts. The issue was in leftover webhook configurations. Our mutating and validating webhook are by design choice owned by `SYSTEM_NAMESPACE` of the particular operator resource, e.g. `knative-serving` or `knative-eventing`. A failed installation might caused by insufficient resources, PDB not being satisfied for all pods etc. 
In those cases in the current impl we won't propagate `status.manifets[]` paths until all deployments are ready. Meanwhile, whenever user tries to delete original CR and re-create new one, there's no cluster-scoped resource deleted by manifests. Manifest are not yet updated in the status, FinalizeKind ends [here](https://github.com/knative/operator/blob/main/pkg/reconciler/knativeserving/knativeserving.go#L90-L92). This leaves cluster-scoped webhook in place, likely already configured without backend deployed. Other resource owned by CR are deleted by garbage collector, but webhook are not if namespace is not deleted by user. 

Subsequent attempts to create in the same namespace, e.g. KnativeServing CR in `knative-serving` that contains ingress class fails on validation webhook /config-map not reachable error. 

I've moved `InstallWebhookConfigs` into `Install` to be applied just after roles are created. That should guarantee any pre-existing webhook configs are rendered inactive. We only apply non-operational [placeholder](https://github.com/knative/operator/blob/main/cmd/operator/kodata/knative-serving/1.18.1/2-serving-core.yaml#L9319-L9345), that's [reconciled](https://github.com/knative/pkg/blob/main/webhook/configmaps/configmaps.go#L118) once webhook is started and connected to validation backend.   

In addition, I've moved `SetManifestPaths` right after `Install` to populate `status.manifests` as soon as resources are appplied to cluster. That should cause cleanup of cluster-resource even for not yet ready installations.

I'll will look into adding e2e tests for leftover webhook case, but meanwhile I'd like to see if tests passes with the changes.

/cc @houshengbo @skonto @maschmid @dprotaso

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving.go
+++ b/pkg/reconciler/knativeserving/knativeserving.go
@@ -88,6 +88,7 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1beta1.Knative
 	}
 
 	if manifest == nil {
+		logger.Warnf("No manifest found; no cluster-scoped resources will be finalized")
 		return nil
 	}
 
@@ -123,10 +124,9 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *v1beta1.KnativeServi
 		r.appendExtensionManifests,
 		r.transform,
 		manifests.Install,
+		manifests.SetManifestPaths, // setting path right after applying manifests to populate paths
 		common.CheckDeployments,
-		common.InstallWebhookConfigs,
 		common.InstallWebhookDependentResources,
-		manifests.SetManifestPaths,
 		common.MarkStatusSuccess,
 		common.DeleteObsoleteResources(ctx, ks, r.installed),
 	}
--- a/pkg/reconciler/common/install.go
+++ b/pkg/reconciler/common/install.go
@@ -55,6 +55,14 @@ func Install(ctx context.Context, manifest *mf.Manifest, instance base.KComponen
 		status.MarkInstallFailed(err.Error())
 		return fmt.Errorf("failed to apply (cluster)rolebindings: %w", err)
 	}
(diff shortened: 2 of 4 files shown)


pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -88,6 +88,7 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1beta1.Knative
 	}
 
 	if manifest == nil {
+		logger.Warnf("No manifest found; no cluster-scoped resources will be finalized")
 		return nil
```
dsimansk, line 92: Adding at a warn here, otherwise it's very silent op. After previous `"Deleting cluster-scoped resources"` there are no more messages in logs, if `status.manifests` are empty.

dsimansk:
My bad I should have added a hold to prevent premature merge, I haven't realized I'm approver in operator repo as well. Anyway I'll create a revert PR, or based on further feedback any additional changes.

PR #2107, opened 2025-06-20
Revert "Propagate manifests paths earlier to operator status (#2106)"

dsimansk:
This reverts commit 2ac3a1eaad81a89738cd7c4d13b6374c0d198b3f.

Per my last comment on #2106. Let's make sure we can revert if needed. Otherwise I'll close based on outcomes of discussion. 

/hold


houshengbo:
@dsimansk I so far did not see anything negatively influencing the operator with the patch merged before. Let's hold on this revert for now, before we find out anything else.

dprotaso:
@dsimansk @houshengbo is there any follow up here?

dprotaso:
@dsimansk @houshengbo is there any follow up here?

PR #2179, opened 2025-10-22
Fix webhook admission control deadlock during installation

### LINKED ISSUE #2178: System TLS Bootsrap: system-internal-tls bootstrap fails with operator: activator crashes before routing-serving-certs secret exists
## What version of Knative?

Knative Operator v1.19.3 (Serving installed via Operator)  
cert-manager v1.18.2  

## Expected Behavior

Bootstrap with `system-internal-tls` enabled should succeed, with Activator starting cleanly once certificates are provisioned.

## Actual Behavior

Applying the `KnativeServing` CR with `system-internal-tls: "Enabled"` causes the Activator to immediately crashloop with the following log:

> Failed to create certificate cache: failed to get activator certificate, secret "routing-serving-certs" not found.
> Enabling system-internal-tls requires the secret to be present and populated with a valid certificate

Restarts of contorller/activator doesn't help, in any order.

Only bootstrapping with `system-internal-tls: "Disabled"` and then patching to `"Enabled"` after the installation works.

## Steps to Reproduce the Problem

Bootstrap Knative Serving (via Operator installation) with `system-internal-tls` enabled on a fresh cluster, where the `routing-serving-certs` Certificate/Secret do not yet exist:

```
apiVersion: v1
kind: Namespace
metadata:
  name: knative-serving
---
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  ingress:
    kourier:
      enabled: true
  config:
    network:
      ingress-class: "kourier.ingress.networking.knative.dev"
      cluster-local-domain-tls: "Enabled"
      system-internal-tls: "Enabled"
    certmanager:
      issuerRef: |
        kind: ClusterIssuer
        name: knative-selfsigned-issuer
      clusterLocalIssuerRef: |
        kind: ClusterIssuer
        name: knative-selfsigned-issuer
      systemInternalIssuerRef: |
        kind: ClusterIssuer
        name: knative-selfsigned-issuer
```

linkvt:
Hi,

this PR fixes a chicken-and-egg bootstrap issue where the operator would get stuck during KnativeServing installation.

Problem:
- ValidatingWebhookConfiguration with failurePolicy=Fail intercepts KCertificate resource creation
- If KCertificate resources are created before the webhook pod is ready, the API server rejects them
- The activator deployment depends on the routing-serving-certs secret (generated from a KCertificate resource which creates a cert-manager Certificate) at runtime
- Previous stage ordering would check all deployments (including activator) before creating KCertificate (and thus cert-manager Certificate) resources, causing a deadlock

Solution:
1. Added CheckWebhookDeployment() that waits specifically for the webhook deployment to be ready before proceeding
2. Reordered reconciliation stages:
   - manifests.Install (creates all deployments and more) - unchanged
   - CheckWebhookDeployment (waits for webhook to be ready) - **new**
   - InstallWebhookDependentResources (creates Certificate resources) - unchanged
   - CheckDeployments (checks all deployments including activator) - unchanged

This ensures:
- Webhook is ready before Certificate creation (avoids admission rejection)
- Certificate resources exist before checking activator
  - `activator` might restart very few times before CM-Certificate was created by `webhook` and Secret with certificate data was created by `cert-manager`

Related functions:
- pkg/reconciler/common/deployments.go: Added CheckWebhookDeployment()
- pkg/reconciler/knativeserving/knativeserving.go: Reordered stages
- pkg/reconciler/common/install.go: Added logging for consistency

Tested locally in my kind cluster with a KnativeServing with and without `system-internal-tls` enabled.

Fixes #2178 

**Release Note**

```release-note
Fix deadlock during creation of KnativeServing with system-internal-tls enabled
```

/kind bug

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving.go
+++ b/pkg/reconciler/knativeserving/knativeserving.go
@@ -124,9 +124,10 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *v1beta1.KnativeServi
 		r.appendExtensionManifests,
 		r.transform,
 		manifests.Install,
-		manifests.SetManifestPaths, // setting path right after applying manifests to populate paths
-		common.CheckDeployments,
+		manifests.SetManifestPaths,    // setting path right after applying manifests to populate paths
+		common.CheckWebhookDeployment, // Wait for webhook to be ready before creating Certificate resources
 		common.InstallWebhookDependentResources,
+		common.CheckDeployments,
 		common.MarkStatusSuccess,
 		common.DeleteObsoleteResources(ctx, ks, r.installed),
 	}
--- a/pkg/reconciler/common/deployments.go
+++ b/pkg/reconciler/common/deployments.go
@@ -19,6 +19,7 @@ package common
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	mf "github.com/manifestival/manifestival"
 	appsv1 "k8s.io/api/apps/v1"
@@ -27,6 +28,7 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 
 	"knative.dev/operator/pkg/apis/operator/base"
+	"knative.dev/pkg/logging"
 )
 
 type deploymentsNotReadyError struct{}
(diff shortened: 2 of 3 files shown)


linkvt:
/remove-approve

Was added automatically as I'm co-release lead.

linkvt:
PTAL @houshengbo 

We're currently preparing release 1.20 so it might be good to get this into this release if you're interested.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -124,9 +124,10 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ks *v1beta1.KnativeServi
 		r.appendExtensionManifests,
 		r.transform,
 		manifests.Install,
-		manifests.SetManifestPaths, // setting path right after applying manifests to populate paths
-		common.CheckDeployments,
+		manifests.SetManifestPaths,    // setting path right after applying manifests to populate paths
+		common.CheckWebhookDeployment, // Wait for webhook to be ready before creating Certificate resources
```
houshengbo: knativeeventing.go also need this function, please add it there as well.
The name of the deployment resource for eventing is "eventing-webhook", you probably need to make the name as a param to the function "CheckWebhookDeployment".

linkvt: Are you sure about eventing needing this as well?

Serving needed it as there are two install steps where the first one blocked the second one (see line 126 and 129):
https://github.com/knative/operator/blob/00cda2a1d39ec2cbaa47bc4fe4471963ecb1d74c/pkg/reconciler/knativeserving/knativeserving.go#L126-L132

Eventing has only one install step which means nothing is broken (see line 143):
https://github.com/knative/operator/blob/00cda2a1d39ec2cbaa47bc4fe4471963ecb1d74c/pkg/reconciler/knativeeventing/knativeeventing.go#L143-L148

houshengbo: Thanks for the explanation!

PR #2267, opened 2026-04-06
Add multi-cluster deployment support via Cluster Inventory API

kahirokunn:
Fixes #2264

Requires https://github.com/knative/infra/pull/827 to be merged first so that the integration-tests-multicluster presubmit exists when this PR's CI runs.

## Proposed Changes

* Enable deploying Knative Serving and Eventing components to remote clusters by setting `spec.clusterProfileRef` on the CR, using the SIG-Multicluster Cluster Inventory API ([KEP-5339](https://github.com/kubernetes/enhancements/issues/5339), `ClusterProfile`) to discover target clusters without depending on a specific fleet manager.
* When `spec.clusterProfileRef` is not set, behavior is completely unchanged — existing single-cluster deployments are unaffected.
* On CR deletion, the operator finalizes resources on the remote cluster. If the remote cluster is unreachable, the finalizer retries; operators can remove it manually to force-delete.

### Design

The core idea is to swap the manifestival `manifest.Client` at the **start** of the reconcile stage pipeline (`ResolveTargetCluster`), so that all subsequent stages — Apply, Delete, Get — transparently operate on the remote cluster.

For garbage collection on the remote cluster, an **anchor ConfigMap** pattern (inspired by [k0smotron](https://github.com/k0smotron/k0smotron)) is used:

- An anchor ConfigMap (`{kind}-{cr-name}-root-owner`) is created on the remote cluster.
- All namespace-scoped resources get an `ownerReference` pointing to this anchor, enabling Kubernetes-native GC.
- Cluster-scoped resources (ClusterRole, etc.) have no `ownerReference` and are explicitly deleted by the finalizer.
- Deleting the anchor ConfigMap cascades to all namespace-scoped resources via GC.

### KnativeServing on a remote cluster

```yaml
kind: KnativeServing
metadata:
  name: knative-serving-apac
  namespace: knative-serving
spec:
  version: "1.21"
  clusterProfileRef:
    name: apac-cluster-01
    namespace: fleet-system
  config:
    network:
      ingress-class: "kourier.ingress.networking.knative.dev"
  ingress:
    kourier:
      enabled: true
```

### KnativeEventing on a remote cluster

```yaml
kind: KnativeEventing
metadata:
  name: knative-eventing-apac
  namespace: knative-eventing
spec:
  version: "1.21"
  clusterProfileRef:
    name: apac-cluster-01
    namespace: fleet-system
```

### Local cluster (unchanged behavior)

```yaml
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  version: "1.21"
# No clusterProfileRef → deploys to the local cluster as before
```

### E2E verification with Kind

> **Note:** The credential plugin used below (`kubeconfig-secretreader-plugin`) is a community plugin.
> An official plugin is being tracked at https://github.com/kubernetes-sigs/cluster-inventory-api/issues/45 —
> once released, switch to the official one.

#### 1. Cluster setup

```bash
# Create Kind clusters
kind create cluster --name hub
kind create cluster --name spoke

# Install ClusterProfile CRD on hub
kubectl --context kind-hub apply -f https://raw.githubusercontent.com/kubernetes-sigs/cluster-inventory-api/main/config/crd/bases/multicluster.x-k8s.io_clusterprofiles.yaml

# Create namespaces on hub
kubectl --context kind-hub create namespace fleet-system
kubectl --context kind-hub create namespace knative-serving
kubectl --context kind-hub create namespace knative-operator
```

#### 2. Spoke cluster credentials

```bash
# Create a ServiceAccount + token Secret on spoke for the operator to use
kubectl --context kind-spoke apply -f - <<EOF
apiVersion: v1
kind: ServiceAccount
metadata:
  name: knative-operator
  namespace: default
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: knative-operator-admin
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: ServiceAccount
  name: knative-operator
  namespace: default
---
apiVersion: v1
kind: Secret
metadata:
  name: knative-operator-token
  namespace: default
  annotations:
    kubernetes.io/service-account.name: knative-operator
type: kubernetes.io/service-account-token
EOF

# Wait for the token to be populated
for i in $(seq 1 30); do
  TOKEN=$(kubectl --context kind-spoke get secret knative-operator-token -o go-template='{{.data.token | base64decode}}' 2>/dev/null) && [ -n "$TOKEN" ] && break
  sleep 1
done

# Copy the token as a Secret on the hub cluster
kubectl --context kind-hub -n knative-operator create secret generic spoke-token \
  --from-literal=token="${TOKEN}" --dry-run=client -o yaml | kubectl --context kind-hub apply -f -
```

#### 3. ClusterProfile

```bash
# Get spoke cluster connection info (Docker-internal IP, not 127.0.0.1)
SPOKE_SERVER="https://$(docker inspect spoke-control-plane --format '{{ .NetworkSettings.Networks.kind.IPAddress }}'):6443"
SPOKE_CA=$(kubectl --context kind-spoke config view --minify --flatten -o jsonpath='{.clusters[0].cluster.certificate-authority-data}')

# Create ClusterProfile (spec only — status is a subresource)
kubectl --context kind-hub apply -f - <<EOF
apiVersion: multicluster.x-k8s.io/v1alpha1
kind: ClusterProfile
metadata:
  name: spoke-cluster
  namespace: fleet-system
spec:
  clusterManager:
    name: kind
EOF

# Patch status with access provider info.
# The extensions tell the credential plugin which Secret to read the token from.
STATUS_PATCH=$(cat <<EOF
{
  "status": {
    "accessProviders": [
      {
        "name": "token-secretreader",
        "cluster": {
          "server": "${SPOKE_SERVER}",
          "certificate-authority-data": "${SPOKE_CA}",
          "extensions": [
            {
              "name": "client.authentication.k8s.io/exec",
              "extension": {
                "secretName": "spoke-token",
                "secretNamespace": "knative-operator",
                "key": "token"
              }
            }
          ]
        }
      }
    ]
  }
}
EOF
)

kubectl --context kind-hub patch clusterprofile spoke-cluster \
  -n fleet-system --type merge --subresource=status -p "${STATUS_PATCH}"
```

#### 4. Operator deployment

```bash
# Build and deploy the operator with ko.
# KIND_CLUSTER_NAME ensures the image is loaded into the correct Kind cluster.
kubectl config use-context kind-hub
KO_DOCKER_REPO=kind.local KIND_CLUSTER_NAME=hub ko apply -f config/

kubectl --context kind-hub -n knative-operator \
  wait --for=condition=Available deployment/knative-operator --timeout=120s

# Create clusterprofile-provider-file ConfigMap (provider name must match accessProviders[].name)
kubectl --context kind-hub -n knative-operator create configmap clusterprofile-provider-file \
  --from-literal=config.json='{"providers":[{"name":"token-secretreader","execConfig":{"apiVersion":"client.authentication.k8s.io/v1","command":"/credential-plugin/kubeconfig-secretreader-plugin","provideClusterInfo":true}}]}'

# Patch the operator deployment:
#   - Mount the credential plugin binary via image volume
#   - Mount the clusterprofile-provider-file config
#   - Add the --clusterprofile-provider-file flag
kubectl --context kind-hub -n knative-operator patch deployment knative-operator --type json -p '[
  {"op":"add","path":"/spec/template/spec/containers/0/args","value":["--clusterprofile-provider-file=/etc/cluster-inventory/config.json"]},
  {"op":"add","path":"/spec/template/spec/containers/0/volumeMounts","value":[
    {"name":"cred-config","mountPath":"/etc/cluster-inventory","readOnly":true},
    {"name":"credential-plugin","mountPath":"/credential-plugin","readOnly":true}
  ]},
  {"op":"add","path":"/spec/template/spec/volumes","value":[
    {"name":"cred-config","configMap":{"name":"clusterprofile-provider-file"}},
    {"name":"credential-plugin","image":{"reference":"ghcr.io/labthrust/kubeconfig-secretreader-plugin:v0.0.1-linux-arm64"}}
  ]}
]'

kubectl --context kind-hub -n knative-operator rollout status deployment/knative-operator --timeout=120s
```

#### 5. Install Envoy Gateway on the spoke cluster

The net-gateway-api ingress provider requires Gateway API resources (GatewayClass, Gateway) to exist on the target cluster **before** deploying KnativeServing. Install [Envoy Gateway](https://gateway.envoyproxy.io/) and create the necessary resources on the spoke cluster.

##### Prerequisites

- helm
- [cloud-provider-kind](https://github.com/kubernetes-sigs/cloud-provider-kind) (for LoadBalancer support in Kind)

```bash
go install sigs.k8s.io/cloud-provider-kind@latest
```

##### Install Envoy Gateway

Since `cloud-provider-kind` installs Gateway API CRDs with server-side apply, `helm install` will fail with field ownership conflicts. Use `helm template` + `kubectl apply --server-side --force-conflicts` to take over ownership.

```bash
cat <<'EOF' > /tmp/values-eg.yaml
config:
  envoyGateway:
    provider:
      type: Kubernetes
      kubernetes:
        deploy:
          type: GatewayNamespace
EOF

export ENVOY_GATEWAY_VERSION=v1.7.1
kubectl --context kind-spoke create namespace envoy-gateway-system

helm template eg oci://docker.io/envoyproxy/gateway-helm \
  --version $ENVOY_GATEWAY_VERSION \
  -n envoy-gateway-system \
  -f /tmp/values-eg.yaml \
  --include-crds | \
  kubectl --context kind-spoke apply --server-side --force-conflicts -f -
```

After installing Envoy Gateway, start cloud-provider-kind in a separate terminal (provides LoadBalancer support for Kind):

```bash
sudo cloud-provider-kind
```

##### Create Gateway API resources (external)

```bash
kubectl --context kind-spoke apply -f - <<EOF
apiVersion: v1
kind: Namespace
metadata:
  name: eg-external
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: knative-external-config
  namespace: envoy-gateway-system
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService:
        name: knative-external
---
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg-external
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: knative-external-config
    namespace: envoy-gateway-system
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg-external
  namespace: eg-external
spec:
  gatewayClassName: eg-external
  listeners:
  - name: http
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All
  - name: tls
    port: 443
    protocol: TLS
    tls:
      mode: Passthrough
    allowedRoutes:
      namespaces:
        from: All
EOF
```

##### Create Gateway API resources (internal)

The internal Gateway uses the `ClusterIP` service type, so it is not accessible from outside the cluster.

```bash
kubectl --context kind-spoke apply -f - <<EOF
apiVersion: v1
kind: Namespace
metadata:
  name: eg-internal
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: knative-internal-config
  namespace: envoy-gateway-system
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService:
        type: ClusterIP
        name: knative-internal
---
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg-internal
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: knative-internal-config
    namespace: envoy-gateway-system
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg-internal
  namespace: eg-internal
spec:
  gatewayClassName: eg-internal
  listeners:
  - name: http
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All
EOF
```

#### 6. Deploy KnativeServing

```bash
kubectl --context kind-hub apply -f - <<EOF
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  clusterProfileRef:
    name: spoke-cluster
    namespace: fleet-system
  ingress:
    gateway-api:
      enabled: true
  config:
    config-gateway:
      external-gateways: |
        - class: eg-external
          gateway: eg-external/eg-external
          service: eg-external/knative-external
          supported-features:
          - HTTPRouteRequestTimeout
      local-gateways: |
        - class: eg-internal
          gateway: eg-internal/eg-internal
          service: eg-internal/knative-internal
          supported-features:
          - HTTPRouteRequestTimeout
    network:
      ingress.class: gateway-api.ingress.networking.knative.dev
    domain:
      example.com: ""
EOF
```

#### 7. Verify

```bash
# Verify resources deployed to spoke cluster
kubectl --context kind-spoke get deployments -n knative-serving
kubectl --context kind-spoke get configmaps -n knative-serving | grep root-owner

# Verify anchor ConfigMap exists with correct labels
kubectl --context kind-spoke get configmap knativeserving-knative-serving-root-owner \
  -n knative-serving -o yaml

# Verify ownerReferences on namespace-scoped resources point to the anchor
kubectl --context kind-spoke get deployment activator -n knative-serving \
  -o jsonpath='{.metadata.ownerReferences}' | jq .

# Deploy a sample application to verify traffic routing
kubectl --context kind-spoke apply -f - <<EOF
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: helloworld-go
  namespace: default
spec:
  template:
    spec:
      containers:
      - image: gcr.io/knative-samples/helloworld-go
        env:
        - name: TARGET
          value: Go Sample v1
EOF

# Get the external gateway IP and test
export LB_IP=$(kubectl --context kind-spoke -n eg-external get svc knative-external \
  -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
echo "$LB_IP helloworld-go.default.example.com" | sudo tee -a /etc/hosts
curl http://helloworld-go.default.example.com

```

#### 8. Verify CR deletion and finalizer cleanup

```bash
# Delete the CR and wait for the finalizer to complete
kubectl --context kind-hub delete knativeserving knative-serving -n knative-serving

# Verify spoke cluster is clean
kubectl --context kind-spoke get all -n knative-serving
kubectl --context kind-spoke get clusterroles | grep knative
```

#### 9. Cleanup

```bash
kind delete cluster --name hub
kind delete cluster --name spoke
```

### Backward compatibility

```bash
# Deploy without --clusterprofile-provider-file and without clusterProfileRef
# → must behave identically to previous releases
kubectl apply -f - <<EOF
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  version: "1.21"
EOF
kubectl get knativeserving -n knative-serving -o jsonpath='{.status.conditions}'
```

**Release Note**

```release-note
Add multi-cluster deployment support via the Cluster Inventory API (KEP-5339). When `spec.clusterProfileRef` is set on a KnativeServing or KnativeEventing CR, the operator deploys components to the remote cluster described by the referenced ClusterProfile. Requires the `--clusterprofile-provider-file` flag to be set on the operator deployment.
```


kahirokunn:
I didn't have the necessary permissions.
`integration-tests-multicluster_operator_main` will keep failing until the following is merged, so I'd like to ignore it using an override until then https://github.com/knative/operator/pull/2267

pkg/apis/operator/v1beta1/knativeserving_lifecycle.go
```diff
@@ -125,6 +126,25 @@ func (is *KnativeServingStatus) MarkDependencyMissing(msg string) {
...
+// MarkTargetClusterNotResolved marks the TargetClusterResolved status as false with the given reason and message.
+func (is *KnativeServingStatus) MarkTargetClusterNotResolved(reason, msg string) {
+	servingCondSet.Manage(is).MarkFalse(
+		base.TargetClusterResolved,
+		reason,
+		msg,
+	)
+	servingCondSet.Manage(is).MarkFalse(
+		base.InstallSucceeded,
+		"TargetClusterUnavailable",
```
houshengbo, line 143: **blocking**: `MarkTargetClusterNotResolved` unconditionally marks `InstallSucceeded=False`.

If the spoke cluster is already running healthy Knative components but the hub temporarily loses connectivity (network blip, ClusterProfile controller restart), this flips `InstallSucceeded` to `False`, which could trigger alerts in monitoring systems. On the next successful reconcile it recovers, but the coupling is overly aggressive.

Consider only marking `InstallSucceeded=False` if install hasn't previously succeeded (i.e., on initial deployment), or introducing a grace period / separate condition that doesn't affect the top-level readiness immediately.

The same applies to `knativeeventing_lifecycle.go`.

houshengbo (commented):
## Review

The PR description is excellent — the verification steps, design explanation, example CRs, and backward compatibility section are thorough and well-written.

**2 blocking items and 7 non-blocking items** are noted in inline comments below.

pkg/reconciler/common/clusterprofile_informer.go
```diff
@@ -0,0 +1,105 @@
...
+	"knative.dev/pkg/logging"
+
+	clusterinventoryv1alpha1 "sigs.k8s.io/cluster-inventory-api/apis/v1alpha1"
+)
+
+const clusterProfileResyncPeriod = 10 * time.Minute
+
+func (c *ClusterProvider) StartInformer(ctx context.Context) {
+	c.mu.Lock()
+	if c.informerStarted {
```
houshengbo, line 37: **blocking**: `clusterprofile_informer.go` has 0% test coverage per Codecov, and `multicluster.go` is at ~54%.

The informer startup path (API availability check, watch/list/handler logic) and the `Refresh`/`doRefresh` error paths in `multicluster.go` are the core of the feature. For a feature that manages remote cluster credentials and client connections, the unit test coverage should be stronger before merging — particularly the error paths in `doRefresh` (ClusterProfile not found, not ready, access provider failure, client creation failure).

pkg/reconciler/common/multicluster.go
```diff
@@ -0,0 +1,615 @@
...
+		entries:       make(map[string]*clusterEntry),
+		access:        accessImpl,
+		ciClient:      ciClient,
+		controllerCtx: controllerCtx,
+		remoteTimeout: defaultRemoteClusterTimeout,
+	}
+
+	go func() {
+		<-controllerCtx.Done()
+		p.CloseAll()
```
houshengbo, line 161: **non-blocking** (low risk): After `controllerCtx.Done()` fires, `CloseAll()` runs. A concurrent reconciler could theoretically call `GetOrRefresh` and re-populate the entries map after cleanup. In practice this is a non-issue since the whole process is shutting down, but adding a `closed` flag that short-circuits `GetOrRefresh` would make the contract explicit.

pkg/reconciler/common/multicluster.go
```diff
@@ -0,0 +1,615 @@
...
+		return entry, "", nil
+	}
+
+	if reason, err := c.Refresh(ctx, namespace, name); err != nil {
+		return nil, reason, err
+	}
+	return c.Get(ctx, key)
+}
+
+func configEqual(a, b *rest.Config) bool {
```
houshengbo, line 345: **non-blocking** (nit): `configEqual` uses `reflect.DeepEqual` for TLS, exec, auth, and impersonate config. This works correctly but if `rest.Config` gains new auth-related fields in a future Kubernetes release, `configEqual` won't detect changes and stale clients could persist. Consider adding a comment noting which fields are intentionally compared, or switching to a hash of the serialized config.

pkg/reconciler/common/multicluster.go
```diff
@@ -0,0 +1,615 @@
...
+}
+
+func AnchorName(instance base.KComponent) string {
+	kind := strings.ToLower(instance.GroupVersionKind().Kind)
+	return kind + "-" + instance.GetName() + "-root-owner"
+}
+
+const maxResourceNameLength = 253
+
+// EnsureAnchorConfigMap creates or updates the anchor ConfigMap on the remote cluster.
```
houshengbo, line 476: **non-blocking**: `EnsureAnchorConfigMap` creates the namespace with only `app.kubernetes.io/managed-by: knative-operator`. If the CR specifies `spec.namespace.labels` or `spec.namespace.annotations` (via `NamespaceConfiguration`), those are applied by `NamespaceConfigurationTransform` on namespace resources in the manifest — but only if the manifest contains a Namespace resource. If it doesn't, the remote namespace will lack those labels. Worth verifying this path or documenting the limitation.

pkg/reconciler/common/transformers.go
```diff
@@ -43,11 +42,18 @@ func transformers(ctx context.Context, obj base.KComponent) []mf.Transformer {
 	}
 }
 
-func injectOwner(owner mf.Owner) mf.Transformer {
+// InjectOwner sets OwnerReference on namespace-scoped resources
+func InjectOwner(owner mf.Owner, anchorOwner mf.Owner) mf.Transformer {
```
houshengbo, line 46: **non-blocking**: If someone accidentally deletes the anchor ConfigMap on the spoke, Kubernetes GC will cascade-delete all namespace-scoped Knative resources. The docs mention this pattern, which is good. Consider making the anchor name more distinctive (e.g., adding a `knative-operator.dev/do-not-delete` annotation) to reduce accidental deletion risk.

pkg/reconciler/common/stages.go
```diff
@@ -28,23 +29,31 @@ import (
 	"knative.dev/operator/pkg/apis/operator/base"
 )
 
+const RemoteDeploymentsPollInterval = 10 * time.Second
```
houshengbo, line 32: **non-blocking**: `RemoteDeploymentsPollInterval = 10s` is hardcoded. With many spoke clusters this produces steady reconcile traffic. Consider exponential backoff for repeated not-ready polls, or making the interval configurable.

test/e2e/knativeserving_spoke_test.go
```diff
@@ -0,0 +1,371 @@
+//go:build e2e && multicluster
```
houshengbo, line 1: **non-blocking**: E2E tests only cover KnativeServing, not KnativeEventing. The eventing reconciler has slightly different finalization logic (the `TLSResourcesPred` handling in `FinalizeKind`). Adding a parallel spoke test for Eventing would strengthen confidence.

kahirokunn: Thanks for the review! Fixed.

pkg/reconciler/common/stages.go
kahirokunn: Thanks for the review! Fixed.

pkg/reconciler/common/transformers.go
kahirokunn: Thanks for the review! Fixed.

pkg/reconciler/common/multicluster.go
kahirokunn: Thanks for the review! Fixed.

pkg/reconciler/common/multicluster.go
kahirokunn: Thanks for the review! Fixed.

pkg/reconciler/common/clusterprofile_informer.go
kahirokunn: Thanks for the review! Fixed.

pkg/apis/operator/v1beta1/knativeserving_lifecycle.go
kahirokunn: Thanks for the review! Fixed.

pkg/reconciler/common/multicluster.go
kahirokunn: Thanks for the review! Fixed.

kahirokunn:
@houshengbo Thank you for your feedback! I have addressed all the points you raised.

kahirokunn:
I will check it

kahirokunn:
All CI passed! 🙌

kahirokunn:
Thank you😆

PR #2287, opened 2026-04-29
Prep for the release of 1.22

houshengbo:


kahirokunn:
Thank you 🙏

kahirokunn:
I found that the multicluster failure is caused by the spoke cluster missing Gateway API CRDs. The `net-gateway-api-controller` deployment is created there, but it does not become ready because the required Gateway API resources are unavailable.

I opened a separate main-based PR with only that fix: https://github.com/knative/operator/pull/2288

kahirokunn:
If you could rebase it on the main branch, I think it will pass CI! 🙌

kahirokunn:
Thank you! Let's go 🙌

kahirokunn:
All CI passed! 🙌

dprotaso:
/hold

still in the process of releasing some repos - eg. kafka-broker

kahirokunn:
Looks like it has been published!

eventing-kafka-broker now has the knative-v1.22.0 release, and the Kafka artifacts are available in GCS as well.

Release:
https://github.com/knative-extensions/eventing-kafka-broker/releases/tag/knative-v1.22.0

I think re-running ./hack/update-codegen.sh should now pick up the missing cmd/operator/kodata/eventing-source/1.22/kafka/... files.

dprotaso:
I don't see the operator needing `eventing-istio` in it's manifests - that still hasn't been released.

@houshengbo can you confirm?

PR #2288, opened 2026-04-30
Install Gateway API CRDs for multicluster tests

kahirokunn:
## What
- Install Gateway API v1.4.1 experimental CRDs on the spoke kind cluster during multicluster e2e setup.
- Wait for the GatewayClass, Gateway, HTTPRoute, and ReferenceGrant CRDs to become Established before wiring the ClusterProfile.

## Why
The multicluster Serving test enables net-gateway-api on the spoke cluster, but the spoke cluster did not install Gateway API CRDs first. With the 1.22 net-gateway-api controller manifest, the controller exposes readiness/liveness probes and stays unavailable when the required Gateway API resources are missing.


kahirokunn:
Thank you 🙏

PR #2306, opened 2026-05-28
Use official secretreader plugin

kahirokunn:
Fixes #

## Background

The multicluster e2e setup previously built and used a test-only exec credential plugin because the official Cluster Inventory API access provider plugins were not available yet. The official `secretreader` and `kubeconfig-secretreader` plugin images have now been released, so this PR switches the examples and tests to use those published images directly.

## Proposed Changes

* Switch the multicluster access provider examples and e2e setup from the test-only self-built exec plugin to the official Cluster Inventory API `secretreader` plugin images.

**Release Note**

```release-note
Multicluster access provider examples and e2e setup now use the official Cluster Inventory API secretreader plugin images.
```


houshengbo (commented):
/lgtm
/approve

kahirokunn:
Thank you 🙏

PR #2314, opened 2026-06-03
Add image transformer for Knative Serving Queue Proxy sidecar

### LINKED ISSUE #2313: Services not using correct image for queue-proxy sidecar
**Describe the bug**

The configuration for overriding the image used by the `queue-proxy` sidecar is not working in my environment.

I am following the documentation to control the images used for Knative serving: https://knative.dev/docs/install/operator/configuring-serving-cr/#download-images-individually-without-secrets

Here is a sanitized version of my `KnativeServing` resource manifest

```yaml
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  registry:
    imagePullSecrets:
    - name: my-secret
    override:
      3scale-kourier-gateway/kourier-gateway: registry.my-org.com/knative-serving-kourier-gateway:1.35.3
      activator: registry.my-org.com/knative-serving-activator:1.22.0
      autoscaler: registry.my-org.com/knative-serving-autoscaler:1.22.0
      autoscaler-hpa: registry.my-org.com/knative-serving-autoscaler-hpa:1.22.0
      controller: registry.my-org.com/knative-serving-controller:1.22.0
      net-kourier-controller/controller: registry.my-org.com/knative-serving-kourier-controller:1.22.0
      queue-proxy: registry.my-org.com/knative-serving-queue:1.22.0
      webhook: registry.my-org.com/knative-serving-webhook:1.22.0
  version: 1.22.0
```

The services continue to use the default image for the `queue-proxy` container. My cluster's `config-deployment` Config Map has the default image specified in it explicitly:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-deployment
  namespace: knative-serving
data:
  queue-sidecar-image: gcr.io/knative-releases/knative.dev/serving/cmd/queue@sha256:371c23206b7bd924474b07e51435198579133eb98fba27a5cb3db9e6ba63b06b
```

**Expected behavior**

My expectation is that this would configure the Services to use my specified image for the `queue-proxy` container.

**To Reproduce**

1. Install Knative Operator v1.22.1
2. Create a `KnativeServing` resource with the `queue-proxy` image overridden
3. Observe Service deployments not using the specified image for the `queue-proxy` container

**Knative release version**

Knative Operator 1.22.1 with Knative Serving 1.22.0

alanrichman:
Fixes #2313

## Proposed Changes

* Adds image transformer for Knative Serving Queue Proxy sidecar

**Release Note**

```release-note
Fix `KnativeServing` registry override not overriding image for Knative Serving Queue Proxy sidecar
```

### DIFF
--- a/pkg/reconciler/common/images.go
+++ b/pkg/reconciler/common/images.go
@@ -43,12 +43,22 @@ var (
 	delimiter             = "/"
 )
 
+const (
+	configMapDeployment  = "config-deployment"
+	queueSidecarImageKey = "queue-sidecar-image"
+	queueProxyOverride   = "queue-proxy"
+)
+
 // ImageTransform updates image with a new registry and tag
 func ImageTransform(registry *base.Registry, log *zap.SugaredLogger) mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
 		// Image resources are handled quite differently, so branch them out early
 		if u.GetKind() == "Image" && u.GetAPIVersion() == "caching.internal.knative.dev/v1alpha1" {
 			return updateCachingImage(registry, u, log)
+		}
+
+		if u.GetKind() == "ConfigMap" && u.GetName() == configMapDeployment {
+			return updateQueueProxyImage(registry, u, log)
 		}
 
 		// Handle all resources that contain a PodSpec.
@@ -180,6 +190,15 @@ func updateCachingImage(registry *base.Registry, u *unstructured.Unstructured, l
 
 	log.Debugw("Finished conversion", "name", u.GetName(), "unstructured", u.Object)
 	return nil
+}
+
+func updateQueueProxyImage(registry *base.Registry, u *unstructured.Unstructured, log *zap.SugaredLogger) error {
+	image, ok := registry.Override[queueProxyOverride]
+	if !ok {
+		return nil
+	}
+	log.Debugw("Updating queue-sidecar-image", "image", image)
+	return unstructured.SetNestedField(u.Object, image, "data", queueSidecarImageKey)
 }
 
 func getImageName(fullImageURL string) string {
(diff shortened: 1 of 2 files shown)


pkg/reconciler/common/images.go
```diff
@@ -182,6 +192,15 @@ func updateCachingImage(registry *base.Registry, u *unstructured.Unstructured, l
 	return nil
 }
 
+func updateQueueProxyImage(registry *base.Registry, u *unstructured.Unstructured, log *zap.SugaredLogger) error {
+	image, ok := registry.Override[queueProxyOverride]
+	if !ok {
+		return nil
```
dsimansk, line 198: In addition to acception targeted value override through direct image ref. I'd say the fix should accept registry level override if present. Simarly to `PodSpec` case a few lines above.[1]

[1] https://github.com/knative/operator/blob/main/pkg/reconciler/common/images.go#L112-L117

dsimansk: @kahirokunn could you pls take a look? I'd appreciate a second opinion here, If it's valuable to address my previous comment. Thanks!

kahirokunn (changes requested):
I left a comment 🙏 
https://github.com/knative/operator/pull/2314#discussion_r3872143383

pkg/reconciler/common/images.go
kahirokunn: Thanks for the ping. I agree this is worth addressing. registry.default should apply to the queue proxy as well, with an explicit queue-proxy override taking precedence.

alanrichman: My need was adequately addressed by @houshengbo's comment here: https://github.com/knative/operator/issues/2313#issuecomment-4670338356

I have been using his recommendation in production for some time with no issues, do you think this PR is worth keeping open and working? I am generally of the opinion that two ways of setting the same end value would probably be confusing. I do still believe that the documentation should be updated to reflect the advice he provided: https://github.com/knative/operator/issues/2313#issuecomment-4670824141

kahirokunn: I think it's fine to continue working on this matter as is.

PR #2316, opened 2026-06-11
Move AggregationRuleTransform to shared common package to fix eventing RBAC race

creydr:
`channelable-manipulator` is a Kubernetes aggregated ClusterRole — its `aggregationRule` tells the
aggregation controller to populate the `rules` field automatically from matching ClusterRoles.

The operator's manifestival reconciliation applies this ClusterRole from its manifest every ~16 seconds.
Because the manifest declares only the `aggregationRule` with `rules: []`, each apply temporarily
clears the aggregated rules. The Kubernetes aggregation controller re-fills them within ~50–200ms,
but during that window any RBAC check against the role returns **403 Forbidden**.

This affects the eventing controller when it patches InMemoryChannels — roughly 0.6% of patch
attempts hit the race window, causing `Subscription` finalizer failures.

The root cause is that `AggregationRuleTransform` (which preserves the cluster's current rules
before applying) was only wired into **KnativeServing** but not **KnativeEventing**.

This PR addresses it by:

* Moving `AggregationRuleTransform` from `knativeserving/common/` to the shared `reconciler/common/` package
* Applying it inside the shared `Transform()` function, so **all** components (Serving, Eventing, Kafka) get it automatically
* Removing the now-redundant call from the KnativeServing transformer chain
* Adding an eventing-specific test case (`channelable-manipulator`) to the existing test

**Release Note**

```release-note
Fix intermittent 403 RBAC errors for aggregated ClusterRoles (e.g. channelable-manipulator) by
preserving Kubernetes-managed rules during reconciliation for all components, not just Serving.
```

### DIFF
--- a/pkg/reconciler/common/aggregated_rules.go
+++ b/pkg/reconciler/common/aggregated_rules.go
@@ -26,12 +26,13 @@ type unstructuredGetter interface {
 	Get(obj *unstructured.Unstructured) (*unstructured.Unstructured, error)
 }
 
-// AggregationRuleTransform
+// AggregationRuleTransform preserves the rules of aggregated ClusterRoles.
+// The Kubernetes aggregation controller manages the rules field of aggregated
+// ClusterRoles. Without this transform, manifestival overwrites the aggregated
+// rules with the empty rules from the manifest, causing a race condition.
 func AggregationRuleTransform(client unstructuredGetter) mf.Transformer {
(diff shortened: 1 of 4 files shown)


aliok (approved):
/lgtm
/approve
/hold

Hold for others

creydr:
@houshengbo can you check on it too?

houshengbo (approved):

PR #2336, opened 2026-08-10
Fix Kourier bootstrap namespace

kahirokunn:
Fixes #

## Proposed Changes

* Use the Kourier bootstrap ConfigMap namespace to replace the default namespace embedded in its Envoy bootstrap configuration.
* Handle the raw ConfigMap data that the common namespace transformer cannot update, following the existing `replaceGatewayNamespace` pattern.
* Add regression coverage for installing Knative Serving outside the `knative-serving` namespace.

**Release Note**

```release-note
Fix Kourier connectivity when Knative Serving is installed in a non-default namespace.
```

### DIFF
--- a/pkg/reconciler/knativeserving/ingress/ingress_test.go
+++ b/pkg/reconciler/knativeserving/ingress/ingress_test.go
@@ -58,7 +58,7 @@ func TestTransformers(t *testing.T) {
 				},
 			},
 		},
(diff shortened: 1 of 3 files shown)


kahirokunn:

dsimansk:
Thanks, looks good! 

/approve
/lgtm

PR #2337, opened 2026-08-10
Add generic resource patches

kahirokunn:
## Summary

This adds a general-purpose `spec.patches` mechanism to `KnativeServing` and `KnativeEventing`.

It covers the long tail of customizations that the Operator does not expose today, without requiring a dedicated API field and transformer for every use case. Patches are applied to the final generated resources, so the same mechanism works for core Serving and Eventing resources, optional Ingress and Source resources, extensions, and `additionalManifests`.

Existing custom resources are unaffected when `spec.patches` is empty.

## Motivation

The Operator already provides typed overrides for common settings. Those remain the preferred interface because they are discoverable and can be validated as part of the API.

However, Kubernetes resources have many fields that are useful only in particular environments. Adding each one to the Operator API would require more API surface, transformation code, generated CRDs, and tests. `spec.patches` provides an escape hatch for those cases while keeping the common paths typed.

## Proposed Changes

- Add `spec.patches` to both `KnativeServing` and `KnativeEventing`.
- Support JSON Patch, JSON Merge Patch, and Strategic Merge Patch.
- Apply patches in declaration order after namespace injection and all existing component transforms.
- Require each target to match exactly one generated resource. `namespace` can disambiguate otherwise identical targets.
- Prevent patches from changing `apiVersion`, `kind`, `metadata.name`, or `metadata.namespace`.
- Support root-level Strategic Merge Patch `$patch: delete` for removing a generated resource from the desired manifest.
- Keep the existing CRD lifecycle behavior by rejecting CRD deletion.

The full usage guide is in [`docs/resource-patches.md`](https://github.com/kahirokunn/knative-operator/blob/resource-patches/docs/resource-patches.md).

## Usage

For example, this changes the rolling update strategy of the Serving controller, which does not need a dedicated Operator API field:

```yaml
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  patches:
  - target:
      apiVersion: apps/v1
      kind: Deployment
      name: controller
    patch:
      type: strategic
      content: |
        spec:
          strategy:
            rollingUpdate:
              maxUnavailable: 0
              maxSurge: 1
```

Targets use the final `apiVersion`, `kind`, `name`, and optional `namespace` of the generated resource. A missing or ambiguous target fails reconciliation so that stale patches are visible during upgrades.

The usage guide also includes an external autoscaler example: deleting a bundled HPA and removing `spec.replicas` from its Deployment lets a controller such as KEDA manage scaling without the Operator restoring those fields.

## Release Note

```release-note
Add spec.patches to KnativeServing and KnativeEventing for customizing generated resources with JSON, merge, or strategic merge patches.
```


kahirokunn:

pkg/reconciler/common/patches_test.go
```diff
@@ -0,0 +1,406 @@
...
+			resources: []unstructured.Unstructured{deploymentForPatchTest(t, "activator", "knative-serving", 1)},
+			patch:     resourcePatchForTest(base.JSONPatchType, "apps/v1", "Deployment", "activator", `not: an-array`),
+			wantError: "decode JSON patch",
+		},
+		{
+			name:      "non-string root strategic directive",
+			resources: []unstructured.Unstructured{deploymentForPatchTest(t, "activator", "knative-serving", 1)},
+			patch:     resourcePatchForTest(base.StrategicMergePatchType, "apps/v1", "Deployment", "activator", `$patch: 1`),
+			wantError: "root $patch directive must be a string",
+		},
```
dsimansk, line 514: Should such test cases produce error when not executed as `strategic` patches? `wantError` is a placeholder text. For me the tests fail on `ApplyResourcePatches() = <nil>`.
```
		{
			name:      "root $patch directive rejected for merge patch",
			resources: []unstructured.Unstructured{deploymentForPatchTest(t, "activator", "knative-serving", 1)},
			patch:     resourcePatchForTest(base.MergePatchType, "apps/v1", "Deployment", "activator", `$patch: delete`),
			wantError: "root $patch directive is only supported for the strategic merge patch type",
		},
		{
			name:      "root $patch directive rejected for json patch",
			resources: []unstructured.Unstructured{deploymentForPatchTest(t, "activator", "knative-serving", 1)},
			patch:     resourcePatchForTest(base.JSONPatchType, "apps/v1", "Deployment", "activator", `$patch: delete`),
			wantError: "root $patch directive is only supported for the strategic merge patch type",
		},
```

dsimansk: If it make sense to add explanation like;

> Resource removal is only available with `type: strategic`. A root-level
`$patch: delete` under a `merge` or `json` patch is rejected rather than applied
as a literal field, so a mistyped patch fails reconciliation instead of silently
leaving the resource in place.

kahirokunn: I've addressed it now. Could you please take another look when you have a chance? Thank you 🙏

kahirokunn: I amended the commit, which made the resulting diff a bit hard to follow. For convenience, I am pasting the changes made specifically for this review comment below:

```diff
diff --git a/docs/resource-patches.md b/docs/resource-patches.md
@@
-The root-level `$patch: delete` directive means the exact HPA identity must
-remain absent. Configure the external autoscaler to use another HPA name.
+Resource removal is available only with `type: strategic`. A root-level
+`$patch: delete` with `type: merge` or `type: json` is rejected rather than
+applied as a literal field, so a mistyped patch fails reconciliation instead of
+silently leaving the resource in place. The directive keeps the exact HPA
+identity absent. Configure the external autoscaler to use another HPA name.

diff --git a/pkg/reconciler/common/patches.go b/pkg/reconciler/common/patches.go
@@
 func isStrategicMergeDelete(patchType base.PatchType, patchJSON []byte) (bool, error) {
-	if patchType != base.StrategicMergePatchType {
-		return false, nil
-	}
 	patchDocument := map[string]interface{}{}
 	if err := json.Unmarshal(patchJSON, &patchDocument); err != nil {
+		if patchType != base.StrategicMergePatchType {
+			return false, nil
+		}
 		return false, fmt.Errorf("decode strategic merge patch: %w", err)
 	}
 	directive, found := patchDocument["$patch"]
 	if !found {
 		return false, nil
 	}
+	if patchType != base.StrategicMergePatchType {
+		return false, errors.New("root $patch directive is only supported for the strategic merge patch type")
+	}
 	directiveString, ok := directive.(string)

diff --git a/pkg/reconciler/common/patches_test.go b/pkg/reconciler/common/patches_test.go
@@
+		{
+			name:      "root $patch directive rejected for merge patch",
+			resources: []unstructured.Unstructured{deploymentForPatchTest(t, "activator", "knative-serving", 1)},
+			patch:     resourcePatchForTest(base.MergePatchType, "apps/v1", "Deployment", "activator", `$patch: delete`),
+			wantError: "root $patch directive is only supported for the strategic merge patch type",
+		},
+		{
+			name:      "root $patch directive rejected for json patch",
+			resources: []unstructured.Unstructured{deploymentForPatchTest(t, "activator", "knative-serving", 1)},
+			patch:     resourcePatchForTest(base.JSONPatchType, "apps/v1", "Deployment", "activator", `$patch: delete`),
+			wantError: "root $patch directive is only supported for the strategic merge patch type",
+		},
```

pkg/reconciler/common/patches.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+// validatePatchDirectives prevents strategic merge directives from becoming
+// literal resource fields when another patch algorithm is selected.
+func validatePatchDirectives(patchType base.PatchType, patchJSON []byte) error {
+	if patchType != base.StrategicMergePatchType {
+		var patchDocument interface{}
+		if err := json.Unmarshal(patchJSON, &patchDocument); err != nil {
+			return nil
+		}
```
dsimansk, line 249: Nit: `json.Unmarshal` error is never returned and silently ignored.

dsimansk:
@kahirokunn flagged one small nit of silent error. Otherwise good to go! Thanks!

pkg/reconciler/common/patches.go
kahirokunn: Thank you. Fixed 🙏

kahirokunn:
@dsimansk Thank you for your review. PTAL again. Thank you 🙏

PR #2338, opened 2026-08-10
Make multicluster CRD waits resilient

kahirokunn:
`kubectl wait --for=condition=Established` can exit immediately when it sees the initial null conditions value. Polling the condition list avoids that race while keeping the existing timeout.

**Release Note**

```release-note
NONE
```


kahirokunn:

PR #2353, opened 2026-09-15
Make operator webhook replicas configurable in Helm

kahirokunn:
Fixes #

## Proposed Changes

* Make the operator-webhook replica count configurable through Helm values.

**Release Note**

```release-note
The operator-webhook replica count is now configurable through Helm values.
```


kahirokunn:
@dsimansk Hi ✋ PTAL. Thank you 🙇

PR #2354, opened 2026-09-16
Log a warning when remote-deployments-poll-interval gets clamped

magic-peach:
This repo has no PR template, so this is a plain factual description.

What changed: added RemoteDeploymentsPollIntervalWasClamped in pkg/reconciler/common/poll_interval_flag.go, and a conditional logger.Warnf call in both pkg/reconciler/knativeserving/controller.go and pkg/reconciler/knativeeventing/controller.go, right next to the existing startup log line that reports the resolved poll interval.

Why: the flag help text and the README's CLI flags table both document that a remote-deployments-poll-interval value below 1s falls back to the default with a warning log entry. The code only clamped the value silently, so a misconfigured flag gave no signal at all. Confirmed via git log that this was an oversight in the PR that introduced the flag (#2267), not something already tracked.

How tested: go build ./... and go test on pkg/reconciler/common, pkg/reconciler/knativeserving, and pkg/reconciler/knativeeventing, all green, including 9 new and existing cases for the poll interval helpers. Rebuilt and retested against the current tip of main immediately before opening this PR.

Confirmed no existing issue or PR covers this (searched by flag name and by keyword).


kahirokunn:
/lgtm
Thank you 🙏

pkg/reconciler/common/poll_interval_flag.go
```diff
@@ -40,3 +40,9 @@ func RemoteDeploymentsPollIntervalValue() time.Duration {
 	}
 	return remoteDeploymentsPollIntervalFlag
 }
```
dsimansk, line 42: I'd rather see bool returned from this function, as it states clamping is happening. Then act accordingly to log warning.

